PR #138, opened 2018-05-23
Docker improvements for local development

danihodovic:
This commit contains various Docker improvements so that Docker and
docker-compose can be used for local development seamlessly.

- volumes are used to mount host source files in the Node container
- Parity is dockerized with docker-compose
- the Node inspector is exposed on 0.0.0.0 to allow debugging from an
  app running within a container

I would like to simplify how configuration is passed to the Node
application, but that's out of the scope of the PR, so I've only
modified it to ensure it works within the Docker environment.

The documentation has been modified to reflect these changes. The
documentation assumes Docker as the first approach to setting up a
development environment. The other approaches were removed.


README.md
```diff
@@ -12,21 +12,32 @@ Scroll to the bottom to get [instructions on how to use postman collection for t
 
 Read below to learn about ambrosus node development.
 
-## Install, setup, tests and linter
-To install dependencies:
+## Running a local development environment using Docker and docker-compose
+
+TODO: Instructions on how to generate the registry address and place it in the
+repository under config
```
danihodovic: This needs to be clarified so that it's a simple as possible to setup a new environment.

danihodovic:
TODO: Modify travis to use a Dockerized test environment so that it's identical to a local development environment.

.travis.yml
```diff
@@ -1,23 +1,22 @@
...
 script:
-  - mongod --version
-  - sleep 2
-  - yarn test
-  - yarn dev:lint
-env:
-  global:
-    - MONGODB_URI="mongodb://localhost/ambrosus_gateway_test"
-    - PORT=9876
+  - docker-compose up -d db parity
```
spherefoundry: you don't really need parity for tests. They run against Ganache;

README.md
spherefoundry: :+1: I think the whole section should be updated:D

docker-compose.yml
```diff
@@ -4,17 +4,33 @@ services:
     build:
       context: .
     container_name: "ambnode"
+    command: yarn dev:start
     environment:
-      - NODE_ENV=production
       - MONGODB_URI=mongodb://db:27017/ambrosus_node
       - MONGODB_DATABASE=ambrosus_node
-      - WEB3_RPC=http://167.99.138.196:8545
+      - WEB3_RPC=http://parity:8545
```
spherefoundry: You have removed support for the WEB3_RPC env variable, so this should also be removed.

test/workers/bundle_downloader.js
```diff
@@ -36,14 +36,14 @@ describe('Bundle downloader - integration', () => {
     web3 = await createWeb3();
   });
 
-  const mongoUri = 'mongodb://localhost:27017/ambrosus_gateway_test2';
+  const mongoUri = 'mongodb://db:27017/ambrosus_gateway_test2';
```
spherefoundry: IMO: this should be taken from the config with only the database name replaced.

danihodovic:
> You have removed support for the WEB3_RPC env variable, so this should also be removed.
Fixed in https://github.com/ambrosus/ambrosus-node/pull/138/commits/fc7980d8fb34ea5ad490a4af2a09b46223346de5

> you don't really need parity for tests. They run against Ganache;
Fixed in https://github.com/ambrosus/ambrosus-node/pull/138/commits/4639bb290ab5534cdb060ec23caa4fc2383635e7

> IMO: this should be taken from the config with only the database name replaced.
Fixed in https://github.com/ambrosus/ambrosus-node/pull/138/commits/684b4d20cfb4874952393f0ccd1911104e2bd1c2 by reading from the configuration and replacing the database name.

danihodovic:
ecf476f changes the context of this PR so that the Node application itself is not Dockerized, but Mongo and Parity are. As agreed in chat discussions this may be an easier approach to Docker rather than full on Dockerizing everything. Especially the Node app is sensitive because it changes developers workflows, the other components less so.

docker-compose.yml
```diff
@@ -1,20 +1,13 @@
...
-    container_name: "ambnode"
-    environment:
-      - NODE_ENV=production
-      - MONGODB_URI=mongodb://db:27017/ambrosus_node
-      - MONGODB_DATABASE=ambrosus_node
-      - WEB3_RPC=http://167.99.138.196:8545
-    ports:
-      - "80:9876"
-    depends_on:
-      - db
```
spherefoundry: Do I understand right that this docker-compose is only meant for development? If this is the case, I think it would be nice to have a comment/disclaimer on top.

danihodovic: Yes, correct. For deployments I'm either going to use a docker file in a separate (private) repository or use Ansible's wrapper around Docker containers (http://docs.ansible.com/ansible/latest/modules/docker_module.html).

I will update with a comment.

PR #141, opened 2018-05-25
Check if Content-Type is application/json

vanruch:


danihodovic:
What is the resulting HTTP status code and body when `throw new InvalidParametersError('application/json MIME type expected');` is called?

vanruch:
@danihodovic just like any error:
```
status: 400,
body: {
    reason: "application/json MIME type expected"
}
```,
it's done in middlewares/error_handling.js

danihodovic:
Cool :+1: 

Could we explicitly specify `The header Content-Type: application/json is required` ?

vanruch:
https://github.com/jshttp/type-is 
It's probably same, but for readability it is a good idea

PR #142, opened 2018-05-25
Read bundle registry contract address from env

danihodovic:
Similar to other config parameters WEB3_NODEPRIVATEKEY, MONGODB_URI,
WEB3_RPC I would like to be able to provide this parameter using the
environment.

The use case is that it simplifies deployments if as much as possible
can be injected using environment variables rather than relying on files
and Docker data volumes.

https://12factor.net/config

### DIFF
--- a/config/dev.js
+++ b/config/dev.js
@@ -1,9 +1,12 @@
 /* eslint-disable header/header */
+
+const registryContractAddress = (
+  process.env.BUNDLE_REGISTRY_CONTRACT_ADDRESS || require('./registryContractAddress.json')
+);
+
 module.exports = {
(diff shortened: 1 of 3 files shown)


danihodovic:
I'll hold on to deploy these changes as @spherefoundry had some ideas on how to improve this.

danihodovic:
@marekkirejczyk I did a test deploy of this commit and it crashed at runtime.

```
/app/node_modules/config/lib/config.js:946
    throw new Error("Cannot parse config file: '" + fullFilename + "': " + e3);
    ^

Error: Cannot parse config file: '/app/config/production.js': Error: Cannot find module './registryContractAddress.json'
```
The workaround is in 00af902.

PR #144, opened 2018-05-25
Find account - filter by accessLevel

vanruch:


apiary.apib
```diff
@@ -320,6 +317,75 @@ Finds newest accounts.
...
+
+    + Body
+
++ Response 403 (application/json)
+
+    Invalid signature or missing register_account permission
+
+    + Body
+
++ Response 404 (application/json)
```
Anze1m: We return 404 on 'get...' endpoints, but on 'find...' endpoint empty array of results is perfectly valid

src/services/account_access_definitions.js
```diff
@@ -60,6 +60,21 @@ export default class AccountAccessDefinitions {
     };
   }
 
+  validateAndCastFindAccountParams(params) {
```
Anze1m: Maybe we should extract this method to some kind of utils or helpers and parametrize it with an array of allowed parameters? We have a very similar method in entity builder for events, and we are also going to create one more, for assets.

src/services/entity_builder.js
```diff
@@ -125,11 +127,8 @@ export default class EntityBuilder {
 
   validateAndCastFindEventsParams(params) {
     const allowedParametersList = ['assetId', 'fromTimestamp', 'toTimestamp', 'page', 'perPage', 'createdBy', 'data'];
-    
-    const invalidFields = Object.keys(params).filter((key) => !allowedParametersList.includes(key));
-    if (invalidFields.length > 0) {
-      throw new InvalidParametersError(`Some parameters (${invalidFields.join(',')}) are not supported`);
-    }
+
+    validateFieldsConstrainedToSet(params, allowedParametersList);
```
Anze1m: 👍

PR #146, opened 2018-05-28
Adding metadata timestamps

Anze1m:


src/services/entity_repository.js
```diff
@@ -143,9 +144,12 @@ export default class EntityRepository {
       'metadata.bundleId': bundleId
     };
 
+    const currentTimestamp = getTimestamp();
+
     const update = {
       $set: {
-        'metadata.bundleTransactionHash': txHash
+        'metadata.bundleTransactionHash': txHash,
+        'metadata.bundleUploadTimestamp': currentTimestamp
```
spherefoundry: I would also add the bundleUploadTimestamp to the bundle itself.  Just 13 lines above:D

PR #149, opened 2018-05-31
Add health checks

danihodovic:
Adds health checks to the application so that we can monitor it's state.
This is important for two reasons:

- On deployment the deployer pings this endpoint to find out if the
  application has booted correctly.

- Once we introduce load balancers the load balancer can detect if the
  application is unhealthy and direct traffic to
  another instance.

### DIFF
--- a/src/routes/health_check.js
+++ b/src/routes/health_check.js
@@ -1,0 +1,32 @@
+/*
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+const healthCheckHandler = (mongoClient, web3) => async (req, res) => {
+  if (!mongoClient.isConnected()) {
+    res.status(500)
+      .type('json')
+      .send({msg: 'Is not connected to MongoDB'});
+    return;
+  }
+
+  try {
+    await web3.eth.getNodeInfo();
+  } catch (err) {
+    res.status(500)
+      .type('json')
+      .send({msg: 'Failed to call getNodeInfo'});
+    return;
+  }
+
+  res.status(200)
+    .type('json')
+    .send();
+};
+
+export default healthCheckHandler;
--- a/src/routes/health_check.js
+++ b/src/routes/health_check.js
@@ -8,25 +8,30 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined 
 */
 
 const healthCheckHandler = (mongoClient, web3) => async (req, res) => {
+  const status = {
+    mongo: {connected: true},
+    web3: {connected: true}
+  };
+
   if (!mongoClient.isConnected()) {
-    res.status(500)
-      .type('json')
-      .send({msg: 'Is not connected to MongoDB'});
-    return;
+    status.mongo.connected = false;
   }
 
   try {
     await web3.eth.getNodeInfo();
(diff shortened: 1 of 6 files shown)


danihodovic:
Do not merge until I can confirm that this works on a testing server. I need a manual confirmation due to the fact that we use ganache in tests and ganache doesn't provide a way of health checking. The `isConnected()` method does not exist, so we stub it into tests.

danihodovic:
```
(node:26) UnhandledPromiseRejectionWarning: TypeError: web3.isConnected is not a function
    at _callee$ (/app/dist/routes/health_check.js:42:24)
    at tryCatch (/app/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:62:40)
```

It turns out that the method that is in the docs doesn't exist https://github.com/ethereum/wiki/wiki/JavaScript-API#web3isconnected. This is understandable because  API doc describe 0.2.x.

I didn't find an `isConnected()` method in the source of `1.0.0-beta.34` so I'll look for a workaround.

danihodovic:
I ended up using `web3.eth.getNodeInfo()` for the eth health check. I ran a test in development where it uses the actual node (not ganache) and disconnected it from the eth node.

It seems to work.
```
curl -i localhost:82/health_check
HTTP/1.1 500 Internal Server Error

{"msg":"Failed to call getNodeInfo"}
```

Mongo health checks are covered in tests.

Health checks for a healthy node also work in a real environment.
```
curl -i localhost:82/health_check
HTTP/1.1 200 OK
```

test/routes/health_check.js
```diff
@@ -0,0 +1,57 @@
...
+
+    expect(response.status).to.eql(500);
+    expect(response.body).to.eql({msg: 'Is not connected to MongoDB'});
+  });
+
+  it('returns 500 if web3 is not connected', async () => {
+    // We can't close the connection to web3 because we're using ganache,
+    // mocking is the second best option.
+    const stub = sinon.stub(apparatus.web3.eth, 'getNodeInfo');
+    stub.returns(Promise.reject('Error: Invalid JSON RPC response: ""'));
```
marekkirejczyk: I think stub needs to be released.

src/server.js
```diff
@@ -17,9 +17,11 @@ import bundlesRouter from './routes/bundles';
 import eventsRouter from './routes/events';
 import tokenRouter from './routes/token';
 import nodeInfoRouter from './routes/nodeinfo';
+import healthCheckHandler from './routes/health_check';
 
 export default class Server {
-  constructor(modelEngine, config) {
+  constructor(modelEngine, mongoClient, config) {
```
marekkirejczyk: I would argue that `mongoClient` should be taken from `modelEngine`.

src/server.js
```diff
@@ -40,6 +42,7 @@ export default class Server {
     app.use('/events', eventsRouter(this.modelEngine.tokenAuthenticator, this.modelEngine.identityManager, this.modelEngine));
     app.use('/token', tokenRouter(this.modelEngine.tokenAuthenticator, this.config));
     app.use('/bundle', bundlesRouter(this.modelEngine));
+    app.use('/health_check', healthCheckHandler(this.mongoClient, this.modelEngine.proofRepository.web3));
```
marekkirejczyk: Great idea about health check. I would argue to use RESTfull naming (i.e. simply `/health`).

src/routes/health_check.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  try {
+    await web3.eth.getNodeInfo();
+  } catch (err) {
+    res.status(500)
+      .type('json')
+      .send({msg: 'Failed to call getNodeInfo'});
+    return;
+  }
```
marekkirejczyk: Do we want to return 500 or json with detailed information, e.g.:
```js
{
mongo: {isConnected: false}, 
nodeInfo: {
  ...
}
}
```

PR #150, opened 2018-05-31
Eslint: disallow trailing whitespace

danihodovic:
Whitespace differences can be picked up by source control systems and flagged as diffs which is annoying. Eslint offers the --fix option that strips any trailing whitespace automatically. If your editor is configured to use eslint you can autoconfigure it to strip trailing spaces.


danihodovic:
Running eslint with `--fix` in your editor automatically fixes this issue.

PR #152, opened 2018-06-07
Validate timestamp as an integer

vanruch:


src/services/entity_builder.js
```diff
@@ -71,6 +71,8 @@ export default class EntityBuilder {
     validateFieldsConstrainedToSet(event, ['content', 'eventId']);
     validateFieldsConstrainedToSet(event.content, ['idData', 'data', 'signature']);
     validateNonNegativeInteger(event.content.idData.accessLevel, `Access level should be a non-negative integer, instead got ${event.content.idData.accessLevel}`);
+    validateNonNegativeInteger(event.content.idData.timestamp, `Timestamp should be a non-negative integer, instead got ${event.content.idData.accessLevel}`);
```
Anze1m:
```
instead got ${event.content.idData.accessLevel}`);
```
I think it rather should be
```
instead got ${event.content.idData.timestamp}`);
```
😉

test/services/entity_builder.js
```diff
@@ -157,9 +157,20 @@ describe('Entity Builder', () => {
...
       });
 
-      it('throws if accessLevel not positive integer', () => {
+      it('throws if timestamp is not a positive integer', () => {
+        let brokenEvent = put('content.idData.timestamp', exampleEvent, '1');
+        expect(() => entityBuilder.validateEvent(brokenEvent)).to.throw(ValidationError);
+        brokenEvent = put('content.idData.timestamp', exampleEvent, 1.1);
+        expect(() => entityBuilder.validateEvent(brokenEvent)).to.throw(ValidationError);
+        brokenEvent = put('content.idData.timestamp', exampleEvent, -1);
+        expect(() => entityBuilder.validateEvent(brokenEvent)).to.throw(ValidationError);
```
Anze1m: I'd add a positive case to see if it does not always throw. Just a sanity check.

test/services/entity_builder.js
```diff
@@ -157,9 +157,20 @@ describe('Entity Builder', () => {
...
+      });
+
+      it('throws if accessLevel is not a positive integer', () => {
         let brokenEvent = put('content.idData.accessLevel', exampleEvent, 1.1);
         expect(() => entityBuilder.validateEvent(brokenEvent)).to.throw(ValidationError);
+        brokenEvent = put('content.idData.accessLevel', exampleEvent, 1.1);
+        expect(() => entityBuilder.validateEvent(brokenEvent)).to.throw(ValidationError);
```
Anze1m: I'd add a positive case to see if it does not always throw. Just a sanity check.

PR #157, opened 2018-07-03
Return latest commit hash in /nodeinfo

vanruch:


apiary.apib
```diff
@@ -988,3 +988,26 @@ Fetches a bundle by identifier
...
+
+## /nodeinfo
+
+#### Get node info [GET]
+
++ Response 200
+
+    + Attributes (object)
+        + commit (string) - Hash of the latest git commit on the node
+        + version (string) - Version of AMB-NET
```
Anze1m: Version of AMB-NET??? Shouldn't it rather be Ambrosus Node?

spherefoundry: true, it should be the version of the node

PR #158, opened 2018-07-03
Specification for basic event entry types.

spherefoundry:


docs/BasicEventEntryTypes.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Asset/Event Identifier
+
+The different identifiers for an asset. Can represent a unique item (e.g. unique serial ID) or a identifier for a type of item (e.g. barcode).
+
+```json
+{
+  *"type": "ambrosus.asset.identifiers",
+  *"identifiers": {
+    "ean13": ["...."],
```
Anze1m: Maybe write `"ean13": ["....", "...."]` instead of `"ean13": ["...."]` to underline the fact, that this arrays can contain more than one item?

PR #161, opened 2018-07-04
Find assets - filter by timestamp

vanruch:


test/services/entity_builder.js
```diff
@@ -404,13 +404,17 @@ describe('Entity Builder', () => {
...
+        toTimestamp: '15123323123'
       };
       const validatedParams = entityBuilder.validateAndCastFindAssetsParams(params);
       expect(validatedParams).to.deep.equal({
         createdBy: anAddress,
         page: 2,
-        perPage: 10
+        perPage: 10,
+        fromTimestamp: 15123123123,
+        toTimestamp: 15123323123
```
Anze1m: Maybe smaller values to make fact that they are different, easier noticeable?

test/services/entity_builder.js
```diff
@@ -430,6 +434,18 @@ describe('Entity Builder', () => {
...
+    it('throws if fromTimestamp is not positive integer', async () => {
+      expect(() => entityBuilder.validateAndCastFindAssetsParams({fromTimestamp: 'avc'})).to.throw(ValidationError);
+      expect(() => entityBuilder.validateAndCastFindAssetsParams({fromTimestamp: '1.3'})).to.throw(ValidationError);
+      expect(() => entityBuilder.validateAndCastFindAssetsParams({fromTimestamp: '-10'})).to.throw(ValidationError);
+    });
+
+    it('throws if toTimestamp is not positive integer', async () => {
+      expect(() => entityBuilder.validateAndCastFindAssetsParams({toTimestamp: 'avc'})).to.throw(ValidationError);
+      expect(() => entityBuilder.validateAndCastFindAssetsParams({toTimestamp: '1.3'})).to.throw(ValidationError);
+      expect(() => entityBuilder.validateAndCastFindAssetsParams({toTimestamp: '-10'})).to.throw(ValidationError);
```
Anze1m: Cases we check: string, non-integer value, negative value. Maybe also add `{}` and `[]`?

vanruch: I think it should be tested in validators test. Testing any possible value for any possible parameter is not possible anyway

test/services/entity_builder.js
vanruch: Wanted to reassemble timestamps, however you may be right

PR #162, opened 2018-07-05
Expose Prometheus metrics

danihodovic:
Exposes Prometheus metrics on the /metrics endpoint.

The prometheus client collects default metrics on Node.js such as heap
and cpu usage.

For a detailed list of default metrics see:
https://github.com/siimon/prom-client/tree/master/lib/metrics

Additionally I've added http request duration metrics which should give
us:

- mean request duration
- median request duration
- request duration quantiles
- number of HTTP requests per second

We can now also alert when ambnode is down by using the Prometheus
metric UP.

### DIFF
--- a/src/middlewares/prometheus_middleware.js
+++ b/src/middlewares/prometheus_middleware.js
@@ -1,0 +1,36 @@
+/*
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+const excludedPaths = ['/health', '/metrics'];
+
+const prometheusMiddleware = (promClient, register) => {
+  const httpRequestDurationSeconds = new promClient.Histogram({
+    register: [register],
+    name: 'http_request_duration_seconds',
+    help: 'Request duration in seconds',
+    buckets: promClient.linearBuckets(0.1, 0.2, 20),
+    labelNames: ['path', 'status']
+  });
+
+  return (req, res, next) => {
+    if (!excludedPaths.includes(req.path)) {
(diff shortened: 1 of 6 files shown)


src/middlewares/prometheus_middleware.js
```diff
@@ -0,0 +1,36 @@
...
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+const excludedPaths = ['/health', '/metrics'];
+
+function prometheusMiddleware(promClient, register) {
```
spherefoundry: We generally prefer to use arrow functions as much as possible. In this `function prometheusMiddleware(promClient, register) {` => `const prometheusMiddleware = (promClient, register) => {`

src/middlewares/prometheus_middleware.js
```diff
@@ -0,0 +1,36 @@
...
+  const httpRequestDurationSeconds = new promClient.Histogram({
+    register: [register],
+    name: 'http_request_duration_seconds',
+    help: 'Request duration in seconds',
+    buckets: promClient.linearBuckets(0.1, 0.2, 20),
+    labelNames: ['path', 'status']
+  });
+
+  return function monitor(req, res, next) {
```
spherefoundry: and here: `return function monitor(req, res, next) {` => `return (req, res, next) => {`

test/integration/prometheus_metrics.js
```diff
@@ -0,0 +1,44 @@
...
+  let apparatus;
+
+  beforeEach(async () => {
+    apparatus = new Apparatus();
+    await apparatus.start();
+  });
+
+  afterEach(async () => {
+    await apparatus.stop();
+  });
```
spherefoundry: :+1: for putting `beforeEach` with `afterEach`

danihodovic:
Addressed the issues, PTAL

danihodovic:
Thanks for the quick review =)

PR #163, opened 2018-07-05
Refactored asset search by identifiers

vanruch:


src/services/data_model_engine.js
```diff
@@ -103,10 +104,25 @@ export default class DataModelEngine {
...
+    return [...new Set(events.results.map((event) => event.content.idData.assetId))];
+  }
+
   async findAssets(params, tokenData) {
-    const validatedParams = this.entityBuilder.validateAndCastFindAssetsParams(params);
-    const accessLevel = await this.accountAccessDefinitions.getTokenCreatorAccessLevel(tokenData);
-    const findAssetQueryObject = this.findAssetQueryObjectFactory.create(validatedParams, accessLevel);
+    let validatedParams = this.entityBuilder.validateAndCastFindAssetsParams(params);
+    if (validatedParams.identifier) {
+      const accessLevel = await this.accountAccessDefinitions.getTokenCreatorAccessLevel(tokenData);
```
Anze1m: Why is this line inside of an `if`?

Anze1m: Why is this line inside of an `if`?

src/services/data_model_engine.js
```diff
@@ -103,10 +104,25 @@ export default class DataModelEngine {
...
+      const consideredAssets = await this.selectAssetsByIdentifier(validatedParams.identifier, accessLevel);
+      validatedParams = pick(put(validatedParams, 'consideredAssets', consideredAssets), 'identifier');
```
spherefoundry: I would change `consideredAssets` to just `assetIds`. In the future we could then introduce direct querying of multiple assets by theirs assetsIds.

src/services/data_model_engine.js
```diff
@@ -103,10 +104,25 @@ export default class DataModelEngine {
     return asset;
   }
 
+  async selectAssetsByIdentifier(identifiers, accessLevel) {
```
spherefoundry: `identifiers` => `identifier` as this may be misunderstood as taking multiple identifiers

spherefoundry: Also the name should be changed to reflect we are returning AssetIDS and not assets directly.

test/services/data_model_engine.js
```diff
@@ -515,6 +503,85 @@ describe('Data Model Engine', () => {
...
+            {content: {idData: {assetId: '10', one: {two: 2}}}},
+            {content: {idData: {assetId: '1'}}}
+          ]
+        });
+        mockAccountAccessDefinitions.getTokenCreatorAccessLevel.resolves(accessLevel);
+
+        modelEngine = new DataModelEngine({}, {}, mockEntityBuilder, {},
+          {}, {}, {}, mockFindEventQueryObjectFactory, {}, mockFindAssetQueryObjectFactory, mockAccountAccessDefinitions);
+
+        result = await modelEngine.selectAssetsByIdentifier(identifier, accessLevel);
```
spherefoundry: you are only using this inside of one it case, so this should move;

test/services/data_model_engine.js
```diff
@@ -515,6 +503,85 @@ describe('Data Model Engine', () => {
...
+        }, accessLevel);
+      });
+
+      it('should execute and return unique ids', () => {
+        expect(mockFindEventQueryObject.execute).to.be.calledOnce;
+        expect(result).to.deep.equal(['1', '10']);
+      });
+
+      it('when accessLevel was not given it defaults to 0', async () => {
+        await modelEngine.selectAssetsByIdentifier(identifier);
```
spherefoundry: Tests for different subjects should be grouped at  least into separate describe blocks

test/services/find_event_query_object.js
```diff
@@ -272,6 +281,51 @@ describe('Find Event Query Object', () => {
...
+
+        it('finds events when select by several ids', async () => {
+          findEventQueryObject = findEventQueryObjectFactory.create(
+            makeAssetIdentifierSearchParams({id: '2', id2: 'abc'}), 100);
+          const found = await findEventQueryObject.execute();
+          expect(found.results).to.deep.equal([eventsSet[2]]);
+          expect(found.resultCount).to.equal(1);
+        });
+
+        it('omits events when the access level in event was higher than user\'s', async () => {
```
spherefoundry: you can skip the `\s` by using \` instead of \' for the whole string

PR #175, opened 2018-07-20
Read git commit info from env

danihodovic:
We run ambnode as a container without git and we don't copy the git
directory to the container. The /nodeInfo endpoint fails for this
reason because it tries to shell out to git and the executable does not
exist in the container.

Instead of adding the git history and installing git in the container
(and shelling out for each request) we can read the git commit as an
environment variable.

At build time of the container we set the build arg to the git commit
and it persists when the container is running.

E.g
```
$ docker build --build-arg GIT_COMMIT=$(git rev-parse --short HEAD) \
	-t ambrosus/ambrosus-node .
```

I also add a label for the git commit so that we can track what version
is running using our monitoring tools.

I've shortened the git commit from the longer version to the commit-ish
which is 7 characters, but this is up for debate and I can easily change
it back.


src/routes/nodeinfo.js
```diff
@@ -9,12 +9,11 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
 import express from 'express';
 import asyncMiddleware from '../middlewares/async_middleware';
-import getGitCommitHash from '../utils/sys_utils';
 
-export const getNodeInfoHandler = (identityManager) => async (req, res) => {
+export const getNodeInfoHandler = (identityManager) => (req, res) => {
   require('pkginfo')(module, 'version');
   res.status(200).send({
-    commit: await getGitCommitHash(),
+    commit: process.env.GIT_COMMIT,
```
spherefoundry: Instead of using process.env.GIT_COMMIT directly, you should provide a method in the Config class that will return you this.

danihodovic: Good idea :+1:

PR #178, opened 2018-07-24
Refactor config

danihodovic:
An attempt to unify the config which is scattered all over the place into a single file. Configuration variables are read from the environment as recommended in https://12factor.net/config.

Once configuration is read, it cannot be altered as the object is frozen.

For test we specify variables in `./test/mocha_env.js` which is read by `./test/mocha.opts` on startup. Most tests use the default mocha config options, but it's possible to change it on a test-by-test basis by providing custom config options to the apparatus. See https://github.com/ambrosus/ambrosus-node/blob/refactor-config/test/helpers/apparatus.js#L25-L30.

For staging / prod they the variables are injected into the environment by the deployment tooling.

For local setup I'd recommend using docker-compose if the process is ran as a container or using [`.envrc`](https://direnv.net/) if it's ran with Node.js on the host.


danihodovic:
The code should be thoroughly reviewed as `src/tasks` have no automated tests AFAIK.

test/helpers/apparatus.js
```diff
@@ -15,23 +15,32 @@ import {cleanDatabase} from '../../src/utils/db_utils';
...
 
 export default class Apparatus extends Application {
   DEFAULT_TOKEN_EXPIRATION = 60 * 60 * 24 * 28;
 
-  constructor() {
+  constructor(customConfig) {
     super(new EmptyLogger());
+    // Read defaults from global config, but allow the options to be customized
+    // for each test.
+    this.config = Object.freeze(Object.assign({}, config, customConfig));
```
spherefoundry: `{...config, ...customConfig}`

config/config.js
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+module.exports = Object.freeze({
```
spherefoundry: :+1: for making it immutable

package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "build": "babel ./src -d ./dist --copy-files",
     "start": "node dist/start.js",
-    "test": "export NODE_ENV=test && mocha --require babel-register --timeout 5000 --recursive",
+    "test": "mocha",
```
spherefoundry: :+1:

src/builder.js
```diff
@@ -30,18 +29,18 @@ class Builder {
     await this.dataModelEngine.addAdminAccount();
   }
 
-  async build(dependencies = {}, config = Config.default()) {
+  async build(dependencies = {}, config) {
```
spherefoundry: I would swap the two parameters or remove the default for the first. Currently to call this method you either `build()` which will give you a `{}, undefined` pair, or `build(dep, config)` which will give you a `dep, config` pair. Which is IMO counterintuitive.

test/middlewares/authorization_header_middleware.js
```diff
@@ -36,12 +36,12 @@ describe('Authorisation header middleware', () => {
   });
 
   describe('Amb authorization is enabled', () => {
-    before(() => {
-      config = {isAuthorizationWithSecretKeyEnabled: () => true};
+    const conf = Object.assign({}, config, {
+      authorizationWithSecretKeyEnabled: true
     });
```
spherefoundry: and here `const conf = {...config, authorizationWithSecretKeyEnabled: true}`

test/mocha.opts
```diff
@@ -0,0 +1,4 @@
+--require babel-register
+--require test/mocha_env.js
+--timeout 5000
+--recursive
```
spherefoundry: :+1:

test/helpers/apparatus.js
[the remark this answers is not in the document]
danihodovic: ES6 elegance :+1:

src/builder.js
danihodovic: Yes that makes sense :+1:

spherefoundry:
+ it would be cool if you would update the README to:D

PR #180, opened 2018-07-25
Mongo auth, ssl and support for replica sets

danihodovic:
Merge https://github.com/ambrosus/ambrosus-node/pull/178 first and rebase on top of that.

The diff will be much smaller once the preceding PR is closed.

This commit adds support for usage of replica sets (RS), SSL and authentication using x509. All three features are opt-in, meaning if you don't provide them they will be omitted when connecting.
I.e
- provide MONGO_X509_USER to enable x509 authentication and SSL
- provide MONGO_REPLICA_SET with the name of the replica set to connect to a RS

x509 authentication is recommended for a production setup. Both the client and the server will use certificates to authenticate and both certificates must be signed by the same certificate authority. SSL is a requirement for this kind of authentication.

- For the client, the servers COMMON NAME in the certificate must match what's provided in the connection string as the host.
- For the server the clients COMMON NAME must match the client certificate and the client must be added to the $external auth database. 

 https://docs.mongodb.com/manual/tutorial/configure-x509-client-authentication/

### DIFF
--- a/src/utils/db_utils.js
+++ b/src/utils/db_utils.js
@@ -8,13 +8,39 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined 
 */
 
 import {MongoClient} from 'mongodb';
-import url from 'url';
+import fs from 'fs';
+import querystring from 'querystring';
 
 const connectToMongo = async (config) => {
-  const uri = config.mongoUri;
-  const database = url.parse(uri).pathname.substr(1);
-  const client = await MongoClient.connect(uri);
-  const db = await client.db(database);
+  const query = {};
+  const options = {};
+  let user = '';
+
+  if (config.mongoReplicaSet) {
+    Object.assign(query, {replicaSet: config.mongoReplicaSet});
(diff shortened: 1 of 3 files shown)


src/utils/db_utils.js
```diff
@@ -8,13 +8,39 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
-  const uri = config.mongoUri;
-  const database = url.parse(uri).pathname.substr(1);
-  const client = await MongoClient.connect(uri);
-  const db = await client.db(database);
+  const query = {};
+  const options = {};
+  let user = '';
+
+  if (config.mongoReplicaSet) {
+    Object.assign(query, {replicaSet: config.mongoReplicaSet});
```
spherefoundry: We have a nice set of methods for that. Please search for `put` for usages.

danihodovic:
Closed in favor of https://github.com/ambrosus/ambrosus-node/pull/218

PR #192, opened 2018-08-02
Check setup against role during startup

vanruch:


src/builder.js
```diff
@@ -21,13 +21,24 @@ import IdentityManager from './services/identity_manager';
...
+import RolesRepository from './services/roles_repository';
 
 class Builder {
   async ensureAdminAccountExist() {
     await this.dataModelEngine.addAdminAccount();
   }
 
+  async ensureAccountIsOnboarded() {
+    const role = await this.rolesRepository.onboardedRole(getDefaultAddress(this.web3));
+    if (role.index === 0) {
```
spherefoundry: Why not `role.name === 'NONE'`?

src/builder.js
```diff
@@ -21,13 +21,24 @@ import IdentityManager from './services/identity_manager';
...
-import {createWeb3} from './utils/web3_tools';
+import {createWeb3, getDefaultAddress} from './utils/web3_tools';
+import RolesRepository from './services/roles_repository';
 
 class Builder {
   async ensureAdminAccountExist() {
     await this.dataModelEngine.addAdminAccount();
   }
 
+  async ensureAccountIsOnboarded() {
```
spherefoundry: This method should accept a list of expected roles, and complain if the actual role is not on it. This will be used as soon as when we implement workers.

src/services/roles_repository.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  get index() {
+    return this.roleIndex;
+  }
+
+  get name() {
+    const roleNames = ['NONE', 'ATLAS', 'HERMES', 'APOLLO'];
+    return roleNames[this.index];
+  }
```
spherefoundry: Bene

test/services/roles_repository.js
```diff
@@ -0,0 +1,55 @@
...
+      })
+    };
+    getRoleStub.returns({
+      call: getRoleCallStub
+    });
+    rolesRepository = new RolesRepository(contractManagerMock);
+  });
+
+  it('calls contract method with correct arguments', async () => {
+    getRoleCallStub.resolves('0');
```
spherefoundry: Something else then `0 ` would be preferred, as `0` is a common default value.

test/services/roles_repository.js
```diff
@@ -0,0 +1,55 @@
...
+    const role = await rolesRepository.onboardedRole(address);
+    expect(getRoleStub).to.be.calledWith(address);
+    expect(getRoleCallStub).to.be.calledOnce;
+    expect(role.name).to.equal('NONE');
+  });
+
+  it('correctly decodes roles names', async () => {
+    getRoleCallStub.resolves('1');
```
spherefoundry: Also `0`

src/tasks/dev_prerun.js
```diff
@@ -38,9 +38,24 @@ async function configureKycContract(dataModelEngine, logger) {
...
+    .send({
+      from: address,
+      gas: 2000000
+    });
+}
+
 async function setupDevelopment(dataModelEngine, logger) {
   await registerAdminAccount(dataModelEngine, logger);
   await configureKycContract(dataModelEngine, logger);
+  await onboardAsHermes(dataModelEngine, logger);
```
Anze1m: Maybe move adding to kyc whitelist to onboarding method, as this actually is part of the onboarding process?
If not then maybe rename `configureKycContract` --> `addToKycWhitelist` or sth.

PR #193, opened 2018-08-02
Disable routes

vanruch:


src/builder.js
```diff
@@ -30,12 +30,12 @@ class Builder {
   }
 
   async ensureAccountIsOnboarded(allowedRoles) {
-    const role = await this.rolesRepository.onboardedRole(getDefaultAddress(this.web3));
-    if (role.name === 'NONE') {
+    this.role = await this.rolesRepository.onboardedRole(getDefaultAddress(this.web3));
+    if (this.role.name === 'NONE') {
```
spherefoundry: `this.role.is(Role.NONE)`?

src/builder.js
```diff
@@ -30,12 +30,12 @@ class Builder {
   }
 
   async ensureAccountIsOnboarded(allowedRoles) {
-    const role = await this.rolesRepository.onboardedRole(getDefaultAddress(this.web3));
-    if (role.name === 'NONE') {
+    this.role = await this.rolesRepository.onboardedRole(getDefaultAddress(this.web3));
```
spherefoundry: Instead of `this.role` I would just return the role the node is onboarded for. Instance variables that are set or not depending on a call to a method is somewhat of an anti-pattern. I know we are doing it already with the build method. But this is not something we should duplicate:D

src/services/roles_repository.js
```diff
@@ -19,17 +19,37 @@ export default class RolesRepository {
...
+  static get ATLAS() {
+    return new Role(1);
+  }
+
+  static get HERMES() {
+    return new Role(2);
+  }
+
+  static get APOLLO() {
+    return new Role(3);
```
spherefoundry: :+1:

PR #194, opened 2018-08-02
Prepare dev:prerun scripts for different node types

vanruch:


src/tasks/dev_prerun.js
```diff
@@ -24,49 +20,15 @@ async function registerAdminAccount(dataModelEngine, logger) {
...
-
-builder.build(config)
-  .then(async ({client, dataModelEngine}) => {
-    await setupDevelopment(dataModelEngine, logger);
-    await client.close();
-  })
-  .catch((exception) => {
-    logger.error(exception);
-    process.exit(1);
-  });
```
Anze1m: Why does this file still exist? Node can be onboarded as Hermes or as Atlas. There is no way to onboard as nobody.
As I understand, this functions are used in `dev_prerun_atlas` and `dev_prerun_hermes`, but in this case they definitely should not be placed in `/tasks`.

src/tasks/dev_prerun_atlas.js
```diff
@@ -0,0 +1,49 @@
...
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+import {WinstonConsoleLogger} from '../utils/loggers';
+import Builder from '../builder';
+import config from '../../config/config';
+import {addToKycWhitelist, registerAdminAccount} from './dev_prerun';
+
+async function onboardAsAtlas(web3, dataModelEngine, logger) {
```
Anze1m: Maybe move this to the same place where `addToKycWhitelist ` and `registerAdminAccount` are?

src/tasks/dev_prerun_hermes.js
```diff
@@ -0,0 +1,47 @@
...
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+import {WinstonConsoleLogger} from '../utils/loggers';
+import Builder from '../builder';
+import config from '../../config/config';
+import {addToKycWhitelist, registerAdminAccount} from './dev_prerun';
+
+async function onboardAsHermes(dataModelEngine, logger) {
```
Anze1m: Maybe move this to the same place where `addToKycWhitelist` and `registerAdminAccount` are?

src/tasks/dev_prerun_atlas.js
vanruch: This one is specific for atlas

src/tasks/dev_prerun_hermes.js
vanruch: This one is specific for hermes

src/tasks/dev_prerun_atlas.js
Anze1m: But it can be used somehere else for test setup purposes.
Task script will be then just gathering proper stuff and running it, without looking into details.

src/tasks/dev_prerun_hermes.js
Anze1m: But it can be used somehere else for test setup purposes.
Task script will be then just gathering proper stuff and running it, without looking into details.

test/integration/bundles.js
```diff
@@ -48,25 +51,8 @@ describe('Bundles - Integrations', () => {
...
-    const rolesContract = await apparatus.contractManager.rolesContract();
-    await rolesContract
-      .methods
-      .onboardAsHermes(url)
-      .send({
-        from: nodeAddress,
-        gas: 2000000
-      });
+    await addToKycWhitelist(Role.HERMES, apparatus.dataModelEngine, new EmptyLogger());
+    await apparatus.rolesRepository.onboardAsHermes(nodeAddress, url);
```
spherefoundry: IMO: we can move this sequence of calls into the apparatus. As we will be using them a lot for all kinds of integration tests.

test/integration/bundles.js
```diff
@@ -46,27 +49,7 @@ describe('Bundles - Integrations', () => {
...
-
-    const rolesContract = await apparatus.contractManager.rolesContract();
-    await rolesContract
-      .methods
-      .onboardAsHermes(url)
-      .send({
-        from: nodeAddress,
-        gas: 2000000
-      });
+    await apparatus.onboardAsHermes(url);
```
spherefoundry: :+1:

PR #195, opened 2018-08-03
Hermes upload worker added

Anze1m:


src/services/upload_terminal.js
```diff
@@ -0,0 +1,30 @@
...
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+import {getDefaultAddress} from '../utils/web3_tools';
+
+
+export default class UploadTerminal {
```
vanruch: I would replace UploadTerminal with 2 repositories (for fees and uploads)

src/start_hermes.js
```diff
@@ -8,10 +8,16 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
 import {WinstonConsoleLogger} from './utils/loggers';
+import config from '../config/config';
+import Builder from './builder';
 import HermesWorker from './workers/hermes_worker';
 
 async function start(logger) {
-  const worker = new HermesWorker(logger);
+  const builder = new Builder();
+  await builder.build(config);
+  await builder.ensureAdminAccountExist();
```
vanruch: Do we need this?

src/workers/hermes_worker.js
```diff
@@ -10,11 +10,18 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+    this.nonce = 0;
   }
 
   async periodicWork() {
-    this.logger.info('Hermes tick');
+    this.logger.info('Trying to upload bundle...');
+    const result = await this.dataModelEngine.finaliseBundle(this.nonce++, this.config.bundleSizeLimit, this.config.storagePeriods);
+    if (result !== null) {
+      this.logger.info('Bundle successfully uploaded');
+    }
```
vanruch: If result==null it means something is wrong and we'd like to have it logged. Additionally it'd be nice to log something like log.info({message: 'Bundle successfully uploaded', bundleId: result.bundleId})

test/helpers/server_apparatus.js
```diff
@@ -37,7 +37,8 @@ export default class ServerApparatus extends Builder {
     const web3 = _web3 || await createWeb3(this.config);
 
     if (!config.headContractAddress) {
-      const headContractAddress = await deployAll(web3, this.logger);
+      const headContract = await deployAll(web3, this.logger);
+      const headContractAddress = headContract.options.address;
```
vanruch: Why did it work before?

test/integration/bundles.js
```diff
@@ -47,9 +46,27 @@ describe.skip('Bundles - Integrations', () => {
...
+
+
+    const rolesContract = await apparatus.contractManager.rolesContract();
+    await rolesContract
+      .methods
+      .onboardAsHermes(url)
+      .send({
+        from,
+        gas: 2000000
+      });
```
vanruch: Can we use repositories for this stuff?

test/services/upload_terminal.js
```diff
@@ -0,0 +1,75 @@
...
+  const getFeeForUploadStub = sinon.stub();
+  const getFeeForUploadCallStub = sinon.stub();
+  const registerBundleStub = sinon.stub();
+  const registerBundleSendStub = sinon.stub();
+  let uploadTerminal;
+  let contractManagerMock;
+  let web3;
+
+  beforeEach(async () => {
+    web3 = await createWeb3();
```
vanruch: Do we need real web3 here?

test/helpers/server_apparatus.js
Anze1m: I didn't.

test/services/upload_terminal.js
Anze1m: I'm afraid that we do 😕

test/integration/bundles.js
Anze1m: I'd use here prerun utils from Your PR.
Will change it after merging Yours.

src/services/upload_terminal.js
Anze1m: I'd leave it like this. We would have about 10 different `ContractRepository` classes.
I would rather see them responsible for whole piece of work, instead of particular contract they interact with.

test/integration/bundles.js
vanruch: https://english.stackexchange.com/questions/30185/you-versus-you-as-polite-form-of-writing

src/services/upload_terminal.js
spherefoundry: I'm with @Anze1m on this. The split into different contracts is a implementation detail of our smart contracts which are a service layer for the node. The interface (called from the model engine) should be very simple here: `upload this stuff, I don't care about the details.` And hence a single provider of this interface is a valid idea.

spherefoundry: Personally, I would prefer `UplaodRepository` as this is a access point to a storage solutions, just as the EntityRepositories are.

src/workers/hermes_worker.js
```diff
@@ -10,11 +10,18 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
 import PeriodicWorker from './periodic_worker';
 
 export default class HermesWorker extends PeriodicWorker {
-  constructor(logger) {
+  constructor(dataModelEngine, config, logger) {
     super(5000, logger);
+    this.dataModelEngine = dataModelEngine;
+    this.config = config;
+    this.nonce = 0;
```
spherefoundry: `BundlingIndex` or `BundleSequenceNumber` is IMO a better name, as `nonce` is somewhat reserved for ethereum transactions.

test/helpers/server_apparatus.js
spherefoundry: :+1:

test/integration/bundles.js
```diff
@@ -47,9 +46,27 @@ describe.skip('Bundles - Integrations', () => {
 
     scenario = new ScenarioBuilder(apparatus.identityManager, apparatusScenarioProcessor(apparatus));
     await scenario.addAdminAccount(adminAccountWithSecret);
-
-    const from = getDefaultAddress(await createWeb3());
-    await apparatus.bundleProofRegistryContract.methods.addToWhitelist(from, url).send({from});
+    const from = apparatus.identityManager.nodeAddress();
```
spherefoundry: Can we please stope using `from` as a name for the sender? I know it makes providing this into the send options easy, but it also makes it really hard to reason about what this is:D

test/integration/bundles.js
```diff
@@ -60,7 +77,7 @@ describe.skip('Bundles - Integrations', () => {
       await scenario.addEvent(0, 1, {timestamp: 2}, [{type: 'e'}])
     ].map(mapEntitiesToIds);
 
-    res = await apparatus.dataModelEngine.finaliseBundle(1);
+    res = await apparatus.dataModelEngine.finaliseBundle(1, undefined, 1);
```
spherefoundry: `16384`?

test/workers/hermes_worker.js
```diff
@@ -0,0 +1,67 @@
...
+      bundleSizeLimit,
+      storagePeriods
+    };
+    mockLogger = {
+      info: sinon.stub()
+    };
+    mockResult = {
+      bundleId: '0xc0ffee'
+    };
+    hermesWorker = new HermesWorker(mockDataModelEngine, mockConfig, mockLogger);
```
spherefoundry: I'm missing a call to `beforeWorkLoop` here, which means the worker may not be fully initialised.

test/workers/hermes_worker.js
```diff
@@ -0,0 +1,72 @@
...
+    hermesWorker.beforeWorkLoop();
```
vanruch: await here

test/workers/hermes_worker.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  afterEach(async () => {
+    hermesWorker.afterWorkLoop();
```
vanruch: await there

PR #197, opened 2018-08-07
Atlas challenge participation worker

vanruch:


src/services/challenges_repository.js
```diff
@@ -0,0 +1,30 @@
...
+  constructor(contractManager) {
+    this.contractManager = contractManager;
+  }
+
+  async previousChallenges(fromBlock = 0) {
+    const challengesContract = await this.contractManager.challengesContract();
+    const events = await challengesContract.getPastEvents('ChallengeCreated', {fromBlock});
+    return {
+      challenges: events.map(({returnValues: {sheltererId, bundleId, challengeId}}) => ({sheltererId, bundleId, challengeId})),
+      maxBlockNumber: events.reduce((maxBlock, {blockNumber}) => Math.max(maxBlock, blockNumber), 0)
```
vanruch: We may want to use getBlock instead but it leaves a chance that getting block and calling getPastEvents happen on different blocks, as a result we either miss some challenges or get them twice

src/workers/atlas_worker.js
```diff
@@ -8,13 +8,38 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+      throw new Error('Not onboarded as ATLAS');
+    }
+    this.nextBlock = await this.web3.eth.getBlockNumber();
+  }
+
+  async tryToResolve({sheltererId, bundleId, challengeId}) {
+    this.logger.info({message: `Trying to fetch the bundle`, sheltererId, bundleId, challengeId});
+    await this.dataModelEngine.downloadBundle(bundleId, sheltererId);
+    await this.challengesRepository.resolveChallenge(challengeId);
+    this.logger.info({message: '🍾 Yahoo! The bundle is ours.', bundleId});
```
Anze1m: We should delete downloaded bundle in case of challenge resolving failure.

src/workers/atlas_worker.js
```diff
@@ -8,13 +8,38 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+  }
+
+  async beforeWorkLoop() {
+    this.logger.info('Atlas worker starting');
+    const role = await this.rolesRepository.onboardedRole(getDefaultAddress(this.web3));
+    if (!role.is(Role.ATLAS)) {
+      this.logger.error('Cannot start apollo worker until being onboarded as APOLLO');
+      throw new Error('Not onboarded as ATLAS');
+    }
+    this.nextBlock = await this.web3.eth.getBlockNumber();
```
Anze1m: Shouldn't we search also for challenges issued before node started?

Anze1m: Btw is this block really "next"? Not more like "current"?

src/workers/atlas_worker.js
```diff
@@ -8,13 +8,47 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+      this.logger.error('Cannot start atlas worker until being onboarded as ATLAS');
+      throw new Error('Not onboarded as ATLAS');
+    }
+    this.nextUnprocessedBlock = await this.web3.eth.getBlockNumber();
+  }
+
+  async tryToResolve({sheltererId, bundleId, challengeId}) {
+    this.logger.info({message: `Trying to fetch the bundle`, sheltererId, bundleId, challengeId});
+    await this.dataModelEngine.downloadBundle(bundleId, sheltererId);
+    await this.challengesRepository.resolveChallenge(challengeId);
```
spherefoundry: We should actually check if the atlas is at the current moment able to resolve it (e.g. it wasn't resolved yet, the atlas is not a shelterer, etc). This is especially important as we expect this rules to be extended shortly. 

IMO: best solution would be to delegate this question to the smart contract. Also: we could check this before downloading, for the first time, and then a second time just before calling resolve.

src/workers/atlas_worker.js
```diff
@@ -8,13 +8,47 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+  }
+
+  async beforeWorkLoop() {
+    this.logger.info('Atlas worker starting');
+    const role = await this.rolesRepository.onboardedRole(getDefaultAddress(this.web3));
+    if (!role.is(Role.ATLAS)) {
+      this.logger.error('Cannot start atlas worker until being onboarded as ATLAS');
+      throw new Error('Not onboarded as ATLAS');
+    }
+    this.nextUnprocessedBlock = await this.web3.eth.getBlockNumber();
```
spherefoundry: This should be set enough in the past to include all challenges started before the worker started, that have still  chance to be resolved (they didn't expire)

src/workers/atlas_worker.js
```diff
@@ -8,13 +8,41 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+    await this.dataModelEngine.downloadBundle(bundleId, sheltererId);
+    await this.challengesRepository.resolveChallenge(challengeId);
+    this.logger.info({message: '🍾 Yahoo! The bundle is ours.', bundleId});
   }
 
   async periodicWork() {
-    this.logger.info('Atlas tick');
+    const challenges = await this.challengesRepository.ongoingChallenges();
+    for (const challenge of challenges) {
+      await this.tryToResolve(challenge).catch((err) => this.logger.error({
```
spherefoundry: you should use try {} catch (err) {} instead of `.catch`

test/services/challenges_repository.js
```diff
@@ -0,0 +1,100 @@
...
+        previousChallenges: sinon.stub().resolves(events),
+        earliestMeaningfulBlock: sinon.stub().resolves(fromBlock),
+        isInProgress: sinon.stub()
+      };
+
+      configWrapperMock = {
+        challengeDuration: sinon.stub().resolves(challengeDuration)
+      };
+
+      challengeWrapperMock.isInProgress.resolves(new Promise((resolve) => resolve(false)));
```
spherefoundry: resolve with a promise?

test/services/contract_wrappers/challenges_wrapper.js
```diff
@@ -0,0 +1,161 @@
...
+    beforeEach(async () => {
+      contractManagerMock = {
+        challengesContract: async () => ({
+          methods: {
+            resolve: resolveChallengeStub
+          }
+        }),
+        web3: {
+          eth: {
+            defaultAccount
```
spherefoundry: this is breaking abstraction. You assume that out getDefaultAddress is using internally web3.eth.defaultAccount. Which may not be true.

test/services/contract_wrappers/config_wrapper.js
```diff
@@ -46,4 +46,30 @@ describe('Config Wrapper', () => {
       expect(atlasStake1CallStub).to.be.calledOnce;
     });
   });
+
+  describe('CHALLENGE_DURATION', () => {
+    const challengeDurationStub = sinon.stub();
+    const challengeDurationCallStub = sinon.stub();
```
spherefoundry: you should recreate the stubs on every `beforeEach` or at least reset them. Currently it only works because there is only one `it`

test/services/contract_wrappers/roles_wrapper.js
```diff
@@ -49,6 +49,33 @@ describe('Roles Wrapper', () => {
     });
   });
 
+  describe('nodeUrl', () => {
+    const nodeUrl = 'url';
+    const getUrlStub = sinon.stub();
```
spherefoundry: Same here

PR #202, opened 2018-08-10
Optimise querying for ongoing challenges

vanruch:


src/services/challenges_repository.js
```diff
@@ -7,19 +7,38 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
-    const challenges = events.map(({returnValues: {sheltererId, bundleId, challengeId}}) => ({sheltererId, bundleId, challengeId}));
-    const isInProgress = await Promise.all(challenges.map(({challengeId}) => this.challengesWrapper.isInProgress(challengeId)));
-    return challenges.filter((challenge, index) => isInProgress[index]);
+    const allChallengeEvents = await this.challengesWrapper.challenges(fromBlock);
+    const resolvedChallengeEvents = await this.challengesWrapper.resolvedChallenges(fromBlock);
+    const timedOutChallengeEvents = await this.challengesWrapper.timedOutChallenges(fromBlock);
+    const notFinishedChallenges = this.filterOutFinished(
+      this.extractChallengeFromEvent(allChallengeEvents),
+      this.extractChallengeFromEvent(resolvedChallengeEvents),
+      this.extractChallengeFromEvent(timedOutChallengeEvents)
```
spherefoundry: All three types of challenges have a different schema. A fact that is ignored completely. I would prefer the `extractChallengeFromEvent` would accept a array of keys we want to extract from the event structure.

src/services/challenges_repository.js
```diff
@@ -7,19 +7,38 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
+    );
+    const isInProgress = await Promise.all(notFinishedChallenges.map(({challengeId}) => this.challengesWrapper.isInProgress(challengeId)));
```
spherefoundry: This is somewhat redundant. As we need to perform this on every single challenge just before we resolve it.

test/services/challenges_repository.js
```diff
@@ -9,49 +9,191 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+      {sheltererId, bundleId, challengeId, count},
+      {sheltererId, bundleId, challengeId: 100, count},
+      {sheltererId, bundleId, challengeId, count}];
+
+    beforeEach(() => {
+      challengesRepository = new ChallengesRepository();
+    });
+
+    it('extracts challengeId, sheltererId, bundleId from events list', async () => {
+      expect(challengesRepository.extractChallengeFromEvent(events)).to.deep.equal(challenges);
```
spherefoundry: This neglects the fact that the method is also used to extract from other types of events.

test/services/challenges_repository.js
```diff
@@ -9,49 +9,191 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+  });
+
   describe('ongoingChallenges', () => {
     const sheltererId = 1;
     const bundleId = 2;
     const challengeId = 3;
     const fromBlock = 4;
+    const count = 1;
     const challengeDuration = 5;
     const events = [
```
spherefoundry: IMO providing this extensive data for this kind of tests is unnecessary. The tests below check if `ongoingChallenges` delegates it's work properly. This placeholders could/should be as small as possible (as to not to introduce noise), just enough to be identified.

src/services/challenges_repository.js
```diff
@@ -7,19 +7,36 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
     const fromBlock = await this.challengesWrapper.earliestMeaningfulBlock(challengeDuration);
-    const events = await this.challengesWrapper.challenges(fromBlock);
-    const challenges = events.map(({returnValues: {sheltererId, bundleId, challengeId}}) => ({sheltererId, bundleId, challengeId}));
-    const isInProgress = await Promise.all(challenges.map(({challengeId}) => this.challengesWrapper.isInProgress(challengeId)));
-    return challenges.filter((challenge, index) => isInProgress[index]);
+    const allChallengeEvents = await this.challengesWrapper.challenges(fromBlock);
+    const resolvedChallengeEvents = await this.challengesWrapper.resolvedChallenges(fromBlock);
+    const timedOutChallengeEvents = await this.challengesWrapper.timedOutChallenges(fromBlock);
+    return this.filterOutFinished(
+      this.extractChallengeFromEvent(allChallengeEvents, ['challengeId', 'sheltererId', 'bundleId', 'count']),
```
spherefoundry: 👍

PR #204, opened 2018-08-13
hermes strategies interface added

Anze1m:


src/services/entity_repository.js
```diff
@@ -135,7 +102,49 @@ export default class EntityRepository {
...
+      .find(thisBundleStubQuery, {fields: this.blacklistedFields})
+      .toArray();
+
+
+    const entities = this.orderedEntityIds(assets, events);
+    const discardedEntities = entities.slice(bundleSizeLimit);
+    await this.unsetEntitiesBundlesStubs(discardedEntities);
+
+    assets = assets.filter((asset) => !discardedEntities.some((id) => asset.assetId === id));
+    events = events.filter((event) => !discardedEntities.some((id) => event.eventId === id));
```
vanruch: Why not `discardedEntities.includes(asset.assetId)`. We can optimize it even more by using Set

src/services/entity_repository.js
```diff
@@ -145,8 +154,21 @@ export default class EntityRepository {
...
+  async revokeEntitiesFromBundling(bundleStubId) {
+    const thisBundleStubQuery = {
+      'repository.bundleStubId': bundleStubId
+    };
+
+    const assets = await this.db.collection('assets').find(thisBundleStubQuery)
+      .toArray();
+    const events = await this.db.collection('events').find(thisBundleStubQuery)
+      .toArray();
+    const entities = this.orderedEntityIds(assets, events);
```
vanruch: There is no need to sort entities. Maybe we can split orderedEntityIds method to map and sort?

src/start_hermes.js
```diff
@@ -17,10 +17,16 @@ async function start(logger) {
...
   await builder.build(config);
   await builder.ensureAccountIsOnboarded([Role.HERMES]);
-  const worker = new HermesWorker(builder.dataModelEngine, config, logger);
+  const strategy = await loadStrategy(config.uploadStrategy);
+  const worker = new HermesWorker(builder.dataModelEngine, config, strategy, logger);
   await worker.start();
 }
 
+async function loadStrategy(uploadStrategy) {
+  const HermesUploadStrategy = require(`./workers/hermes_strategies/${uploadStrategy}`).default;
```
vanruch: Let's maybe keep whole path in config

src/workers/hermes_strategies/regular_interval_upload_strategy.js
```diff
@@ -0,0 +1,34 @@
...
+import {getTimestamp} from '../../utils/time_utils';
+
+export default class RegularIntervalUploadStrategy extends HermesUploadStrategy {
+  constructor() {
+    super();
+    this.lastBundlingTimestamp = null;
+  }
+
+  bundlingPeriod() {
+    return 2000;
```
vanruch: Isn't 2000 too small with block time 5 seconds?

src/workers/hermes_strategies/upload_strategy.js
```diff
@@ -0,0 +1,26 @@
...
+
+  storagePeriods() {
+    return 2;
+  }
+
+  async shouldBundle() {
+    throw new Error('Should be implemented');
+  }
+
+  async didBundle() {
```
vanruch: afterBundling?

src/workers/hermes_strategies/upload_strategy.js
```diff
@@ -0,0 +1,26 @@
...
+export default class HermesUploadStrategy {
+  bundlingPeriod() {
+    return 1000;
+  }
+
+  storagePeriods() {
+    return 2;
+  }
+
+  async shouldBundle() {
```
vanruch: decideIfShouldBundle?

src/start_hermes.js
Anze1m: Putting all strategies in one directory requires some discipline when managing node. Reduces mess and simplifies config. I'd leave it like this.

src/workers/hermes_strategies/upload_strategy.js
```diff
@@ -0,0 +1,26 @@
...
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+export default class HermesUploadStrategy {
+  bundlingPeriod() {
```
vanruch: Let's not use period here to not confuse with storagePeriods. bundlingInterval seems more appropriate

src/workers/hermes_strategies/regular_interval_upload_strategy.js
Anze1m: The consequence would be the fact that there will be more than one bundle added during one block. Bundling frequency should be dependent on traffic generated on node, not on the blockchain details. Also this is only sample strategy, so I haven't really analysed what is correct in business way and what is not. Specific strategies to be added in another ticket(s)

src/workers/hermes_strategies/upload_strategy.js
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+export default class HermesUploadStrategy {
```
vanruch: Imagine scenario: I want to check fee and then decide if I want to bundle.
Idea 1: inject dataModelEngine to the strategy
Idea 2: make method desiredComponents or similar with component names and inject them from the builder

src/start_hermes.js
vanruch: If someone starts it from docker as a black box it requires some black magick to put there

src/workers/hermes_strategies/upload_strategy.js
Anze1m: Wowowow... Does not look beautiful, but maybe it is needed...
@spherefoundry ???

src/workers/hermes_strategies/upload_strategy.js
Anze1m: Imo shouldBundle is ok. Simple descriptive question and yes/no answer.

src/workers/hermes_strategies/upload_strategy.js
Anze1m: my proposal is `bundlingSucceeded`

src/workers/hermes_worker.js
```diff
@@ -8,20 +8,39 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
-    super(5000, logger);
+  constructor(dataModelEngine, config, strategy, logger) {
+    super(strategy.bundlingPeriod(), logger);
     this.dataModelEngine = dataModelEngine;
-    this.config = config;
     this.bundleSequenceNumber = 0;
+    this.strategy = strategy;
+    this.config = config;
+    if (!(this.strategy instanceof HermesUploadStrategy)) {
+      throw new Error('A valid strategy must provided');
```
vanruch: must be provided

src/services/entity_repository.js
Anze1m: Propose exact implementation, because I'm not sure if we can really do it better.

src/start_hermes.js
```diff
@@ -17,10 +17,16 @@ async function start(logger) {
   const builder = new Builder();
   await builder.build(config);
   await builder.ensureAccountIsOnboarded([Role.HERMES]);
-  const worker = new HermesWorker(builder.dataModelEngine, config, logger);
+  const strategy = await loadStrategy(config.uploadStrategy);
+  const worker = new HermesWorker(builder.dataModelEngine, config, strategy, logger);
   await worker.start();
 }
 
+async function loadStrategy(uploadStrategy) {
```
vanruch: why is this async?

src/services/entity_repository.js
```diff
@@ -144,9 +157,22 @@ export default class EntityRepository {
...
     const events = await this.db.collection('events').find(thisBundleStubQuery)
       .toArray();
-    const entities = this.orderedEntityIds(assets, events);
-    await this.setEntitiesBundles(entities.slice(0, bundleSizeLimit), bundleId);
-    await this.unsetEntitiesBundlesStubs(entities.slice(bundleSizeLimit), bundleId);
+    const entities = this.assetsAndEventsToEntityIds(assets, events);
+    await this.setEntitiesBundles(entities, bundleId);
+  }
+
+  async revokeEntitiesFromBundling(bundleStubId) {
```
spherefoundry: `discardBundling`?

test/workers/hermes_worker.js
```diff
@@ -19,54 +20,104 @@ describe('Hermes Worker', () => {
...
+    it('increments bundleSequenceNumber', async () => {
+      const bundleSequenceNumberAfter = hermesWorker.bundleSequenceNumber;
+      expect(bundleSequenceNumberAfter - bundleSequenceNumber).to.equal(1);
+    });
+
+    it('informs strategy', async () => {
+      expect(bundlingSucceededStub).to.have.been.calledOnce;
+    });
+
+    it('uses logger twice', async () => {
```
spherefoundry: Does it make sense if we really shouldn't test if logger was used?

PR #205, opened 2018-08-16
Atlas challenge participation strategy interface

vanruch:


src/workers/atlas_strategies/challenge_resolution_strategy.js
```diff
@@ -0,0 +1,25 @@
...
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+/** @abstract */
+export default class ChallengeResolutionStrategy {
+  get challengePullingInterval() {
```
spherefoundry: I would stick with `workerInterval` as in the Hermes upload strategy. 

On the other hand 👍 for using a property instead of a function. Could you update the Hermes strategy to also use one?

src/workers/atlas_worker.js
```diff
@@ -8,20 +8,38 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+      this.logger.info({message: 'Challenge resolution cancelled', challengeId});
+    }
   }
 
   async tryToResolve({sheltererId, bundleId, challengeId}) {
-    this.logger.info({message: `Trying to fetch the bundle`, sheltererId, bundleId, challengeId});
-    await this.dataModelEngine.downloadBundle(bundleId, sheltererId);
-    await this.challengesRepository.resolveChallenge(challengeId);
-    this.logger.info({message: '🍾 Yahoo! The bundle is ours.', bundleId});
+    if (await this.strategy.shouldFetchBundle({sheltererId, bundleId, challengeId})) {
```
spherefoundry: You provide here parameters that are not declared in the interface of the ChallengeResolutionStrategy.

src/workers/atlas_strategies/challenge_resolution_strategy.js
spherefoundry: Also: the name of the class should be `AtlasChallengeParticipationStrategy`

src/workers/atlas_worker.js
```diff
@@ -8,20 +8,38 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
     this.dataModelEngine = dataModelEngine;
+    this.strategy = strategy;
     this.challengesRepository = challengesRepository;
+    if (!(this.strategy instanceof ChallengeResolutionStrategy)) {
+      throw new Error('A valid strategy must be provided');
+    }
+  }
+
+  async resolve(bundle, challengeId) {
+    if (await this.strategy.shouldResolveChallenge(bundle)) {
```
spherefoundry: I would change the way the code is organised here. We currently have a call chain where `periodicWork` is calling `tryResolve` which is calling `resolve`. In the two later cases the the names are extremely confusing. The first is actually consulting the strategy is to try to download the bundle, and then tries to download. The second is consulting the strategy is to resolve a downloaded bundle and then resolves. I would rather move all the strategy consulting into `periodicWork` and from there call `tryToDownload` and `tryToResolve`

src/workers/atlas_worker.js
vanruch: Because if we declare arguments in abstract method the linter becomes angry

PR #208, opened 2018-08-17
Add link to docs at unknown paths

vanruch:


test/integration/fallback.js
```diff
@@ -0,0 +1,53 @@
...
+chai.use(properAddress);
+chai.use(chaiHttp);
+chai.use(chaiAsPromised);
+
+const {expect} = chai;
+
+describe('Fallback - Integrations', async () => {
+  let apparatus;
+
+  describe('Check if it works', async () => {
```
spherefoundry: Adding this describe level doesn't make sense. Just put the code into the top level one.

PR #212, opened 2018-08-20
bundle size validation added

Anze1m:


src/services/data_model_engine.js
```diff
@@ -204,8 +204,17 @@ export default class DataModelEngine {
...
     await this.entityRepository.storeBundle(bundle);
     return bundle;
   }
+
+  async verifyBundle(bundle) {
+    const bundleSizeLimit = await this.contractManager.configWrapper.bundleSizeLimit();
+    if (bundle.content.entries.length > bundleSizeLimit) {
+      throw new Error('Bundle size surpasses the limit');
+    }
+    // TODO another on-chain verifications
```
vanruch: IMHO we don't need todo here as we haven't decided what we want to do :)

src/services/data_model_engine.js
```diff
@@ -204,8 +204,17 @@ export default class DataModelEngine {
...
     }
-    await this.entityBuilder.validateBundle(bundle);
+    this.entityBuilder.validateBundle(bundle);
+    await this.verifyBundle(bundle);
     await this.entityRepository.storeBundle(bundle);
     return bundle;
   }
+
+  async verifyBundle(bundle) {
+    const bundleSizeLimit = await this.contractManager.configWrapper.bundleSizeLimit();
```
vanruch: Why using wrappers outside repositories?

src/services/data_model_engine.js
```diff
@@ -204,8 +204,17 @@ export default class DataModelEngine {
...
+    if (bundle.content.entries.length > bundleSizeLimit) {
+      throw new Error('Bundle size surpasses the limit');
```
vanruch: ValidationError

test/services/data_model_engine.js
```diff
@@ -1135,6 +1136,12 @@ describe('Data Model Engine', () => {
         entityBuilder: mockEntityBuilder,
         rolesRepository: mockRolesRepository
       });
+
+      verifyBundleStub = sinon.stub(modelEngine, 'verifyBundle').resolves();
+    });
+
+    afterEach(() => {
+      verifyBundleStub.restore();
```
vanruch: I think restoring stub on object that we create in beforeEach isn't needed, though i'm not sure

test/services/data_model_engine.js
```diff
@@ -1152,9 +1159,43 @@ describe('Data Model Engine', () => {
...
+
+    it('does not store bundle if on-chain verification failed', async () => {
+      verifyBundleStub.rejects();
+      await expect(modelEngine.downloadBundle(bundleId, sheltererId)).to.be.rejected;
+      expect(mockEntityRepository.storeBundle).to.be.not.called;
+    });
+  });
+
+  describe('Verifying a bundle', async () => {
+    const downloadedBundle = createBundle({},['first', 'second', 'third']);
```
vanruch: Missed space after `,`. Should check why linter did not fail

src/services/data_model_engine.js
vanruch: Btw we have it in hermes worker too

test/services/data_model_engine.js
Anze1m: Just in case

test/services/data_model_engine.js
Anze1m: 👍

src/services/data_model_engine.js
Anze1m: We have more or less decided on validating metadata, but probably ur right.

src/services/data_model_engine.js
Anze1m: 👍

src/services/data_model_engine.js
Anze1m: Config contract has only getters. Not sure if need another class just because we want name like `configRepository` instead of `configWrapper`. We probably should forward all call to wrappers, through repositories, but in this particular case I'm not so confident about it.

PR #215, opened 2018-08-21
Refactor: Simplify access token middleware

danihodovic:


src/middlewares/access_token_middleware.js
```diff
@@ -11,24 +11,19 @@ import {ValidationError, AuthenticationError} from '../errors/errors';
...
   const [type, token] = authorizationHeader.split(' ');
   if (type !== 'AMB_TOKEN') {
-    if (required) {
-      throw new ValidationError(`Expected Authorization type AMB_TOKEN`);
-    } else {
-      next();
-      return;
-    }
+    throw new ValidationError(`Expected Authorization type AMB_TOKEN`);
   }
```
spherefoundry: This is not the same code:D The original allowed for a token to be passed in optionally (when required was set to false). In you version it will be discarded.

PR #216, opened 2018-08-22
Request size limit

vanruch:
Throw 413 on too large requests

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -31,5 +31,5 @@ module.exports = Object.freeze({
 
   docsLink: process.env.DOCS_LINK || 'https://dev.ambrosus.com/',
 
-  requestSizeLimit: process.env.REQUEST_SIZE_LIMIT || '5mb'
+  requestSizeLimit: process.env.REQUEST_SIZE_LIMIT || '1mb'
 });
--- a/config/config.js
+++ b/config/config.js
@@ -29,5 +29,7 @@ module.exports = Object.freeze({
 
   gitCommit: process.env.GIT_COMMIT,
 
-  docsLink: process.env.DOCS_LINK || 'https://dev.ambrosus.com/'
+  docsLink: process.env.DOCS_LINK || 'https://dev.ambrosus.com/',
+
+  requestSizeLimit: process.env.REQUEST_SIZE_LIMIT || '5mb'
 });
--- a/src/middlewares/error_handling.js
+++ b/src/middlewares/error_handling.js
@@ -16,8 +16,7 @@ import {
 
 export default (logger) => (err, req, res, next) => {
   let status;
-
-  if (err instanceof ValidationError) {
+  if (err instanceof ValidationError || err.type === 'entity.parse.failed') {
     status = 400;
   } else if (err instanceof AuthenticationError) {
     status = 401;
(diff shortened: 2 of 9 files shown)


config/config.js
```diff
@@ -29,5 +29,7 @@ module.exports = Object.freeze({
 
   gitCommit: process.env.GIT_COMMIT,
 
-  docsLink: process.env.DOCS_LINK || 'https://dev.ambrosus.com/'
+  docsLink: process.env.DOCS_LINK || 'https://dev.ambrosus.com/',
+
+  requestSizeLimit: process.env.REQUEST_SIZE_LIMIT || '5mb'
```
Anze1m: @danihodovic at first proposed 10mb. Maybe stick to this as a default value?

src/middlewares/error_handling.js
```diff
@@ -25,6 +25,8 @@ export default (logger) => (err, req, res, next) => {
     status = 403;
   } else if (err instanceof NotFoundError) {
     status = 404;
+  } else if (err.type === 'entity.too.large') {
+    status = 413;
```
Anze1m: Shouldn’t we also add 406 for wrong content type?

test/services/upload_repository.js
```diff
@@ -62,7 +62,7 @@ describe('Upload repository', () => {
       await expect(uploadRepository.verifyBundle(downloadedBundle)).to.be.fulfilled;
     });
 
-    it('throws if downloaded bundle is to big', async () => {
+    it('throws if downloaded bundle is too big', async () => {
```
Anze1m: 😻

config/config.js
danihodovic: I'm fine with either. We can modify it down the road.

test/integration/max_request_limit.js
```diff
@@ -0,0 +1,60 @@
...
+    await expect(request).to.be.eventually.rejected.and.have.property('status', 413);
+  });
+
+  it('should accept requests under requestSizeLimit', async () => {
+    const data = generateStringOfSize(requestSizeLimit);
+    const request = apparatus.request()
+      .post('/accounts')
+      .type('json')
+      .send(data);
+    await expect(request).to.be.eventually.rejected.and.have.property('status', 500);
```
spherefoundry: A) What a particularly interesting way to check an expectation. I would take something that actually accepts some data, send it, and await a 200
B) If you find a endpoint that returns `500` for whatever reason. Please feel obligated to fix it right away 😛

config/config.js
danihodovic: Actually, can we stick to 1mb?

src/middlewares/error_handling.js
```diff
@@ -16,15 +16,16 @@ import {
...
+  if (err instanceof ValidationError || err.type === 'entity.parse.failed') {
     status = 400;
   } else if (err instanceof AuthenticationError) {
     status = 401;
   } else if (err instanceof PermissionError) {
     status = 403;
   } else if (err instanceof NotFoundError) {
     status = 404;
+  } else if (err.type === 'entity.too.large') {
+    status = 413;
```
danihodovic: Do we need to change the status code here? Body parser seems to return 413 when the payload is too big: https://github.com/expressjs/body-parser/blob/e6ccf98015fece0851c0c673fc2776c30ad79e5d/test/json.js#L113-L121

We could add a test case to try this.

vanruch: It looks like this middleware catches this error and as the type is unknown, the 500 is thrown

danihodovic: OK :+1:

PR #226, opened 2018-08-30
Hermes worker - check funds are available before trying to bundle

vanruch:
\+ got rid of direct usage of contract wrappers in hermes worker

### DIFF
--- a/src/services/contract_wrappers/fees_wrapper.js
+++ b/src/services/contract_wrappers/fees_wrapper.js
@@ -8,6 +8,7 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined 
 */
 
 import ContractWrapper from './contract_wrapper';
+import BN from 'bn.js';
 
 export default class FeesWrapper extends ContractWrapper {
   async contract() {
@@ -17,5 +18,10 @@ export default class FeesWrapper extends ContractWrapper {
   async feeForUpload(storagePeriods) {
     const contract = await this.contract();
     return contract.methods.getFeeForUpload(storagePeriods).call();
+  }
+
+  async checkIfEnoughFunds(fee) {
+    const balance = new BN(await this.web3.eth.getBalance(this.contractManager.defaultAddress()));
+    return balance.gte(new BN(fee));
   }
 }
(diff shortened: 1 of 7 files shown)


src/services/contract_wrappers/fees_wrapper.js
```diff
@@ -18,4 +19,11 @@ export default class FeesWrapper extends ContractWrapper {
     const contract = await this.contract();
     return contract.methods.getFeeForUpload(storagePeriods).call();
   }
+
+  async ensureEnoughFunds(fee) {
+    const balance = new BN(await this.web3.eth.getBalance(this.contractManager.defaultAddress()));
+    if (balance.lt(new BN(fee))) {
+      throw new Error(`Insufficient funds: required: ${fee} and got ${balance}`);
+    }
+  }
```
Anze1m: IMO too much logic for a wrapper. Let wrapper be wrapper and move it out somewhere else. Maybe to `uploadRepository`? Also I'd make direct call to fees (obvsly through `feesWrapper`) to get current fee, instead of passing it to method.

src/services/upload_repository.js
```diff
@@ -20,6 +20,7 @@ export default class UploadRepository {
 
   async uploadBundle(bundleId, storagePeriods) {
     const fee = await this.feesWrapper.feeForUpload(storagePeriods);
+    await this.feesWrapper.ensureEnoughFunds(fee);
```
Anze1m: Yeah... You get `fee` from `feesWrapper` just to pass it again to `feesWrapper`, to check if it is enough.

src/services/upload_repository.js
```diff
@@ -31,8 +32,12 @@ export default class UploadRepository {
     return this.shelteringWrapper.shelteringExpirationDate(bundleId);
   }
 
+  async bundleSizeLimit() {
+    return this.configWrapper.bundleSizeLimit();
+  }
+
   async verifyBundle(bundle) {
-    const bundleSizeLimit = await this.configWrapper.bundleSizeLimit();
+    const bundleSizeLimit = await this.bundleSizeLimit();
```
Anze1m: ❓❓❓
Why not call through `configWrapper`? 😜

src/services/contract_wrappers/fees_wrapper.js
vanruch: i wish i could but we use wrappers to separate web3 from rest of the world, so it's either this or put a getBalance method to some wrapper

PR #234, opened 2018-09-13
Add health check for Hermes / Atlas workers

danihodovic:
Adds a health checks so that we can monitor that the worker is healthy. Used after deploys and for periodical probes.

### DIFF
--- a/src/workers/hermes_worker.js
+++ b/src/workers/hermes_worker.js
@@ -7,17 +7,33 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v. 
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 
+import express from 'express';
+import asyncMiddleware from '../middlewares/async_middleware';
+import healthCheckHandler from '../routes/health_check';
 import PeriodicWorker from './periodic_worker';
 import HermesUploadStrategy from './hermes_strategies/upload_strategy';
 
-
 export default class HermesWorker extends PeriodicWorker {
-  constructor(dataModelEngine, workerLogRepository, strategy, logger) {
+  constructor(
+    dataModelEngine,
+    workerLogRepository,
+    strategy,
+    logger,
+    mongoClient,
+    serverPort
+  ) {
     super(strategy.workerInterval, logger);
     this.dataModelEngine = dataModelEngine;
     this.bundleSequenceNumber = 0;
     this.strategy = strategy;
     this.workerLogRepository = workerLogRepository;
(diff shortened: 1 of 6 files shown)


src/workers/hermes_worker.js
```diff
@@ -7,21 +7,46 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
+
     if (!(this.strategy instanceof HermesUploadStrategy)) {
       throw new Error('A valid strategy must be provided');
     }
   }
 
+  async work() {
+    // .call() is a workaround for
+    // https://github.com/babel/babel/issues/3930#issuecomment-254827874
+    await super.work.call(this);
```
spherefoundry: Instead of overriding work and teardown, you should use the `beforeWorkLoop` and `afterWorkLoop` methods.

src/workers/hermes_worker.js
```diff
@@ -7,21 +7,46 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
     this.bundleSequenceNumber = 0;
     this.strategy = strategy;
     this.uploadRepository = uploadRepository;
+    this.expressApp = express();
+    this.serverPort = serverPort;
+
+    this.expressApp.get('/health', asyncMiddleware(
+      healthCheckHandler(dataModelEngine.mongoClient, uploadRepository.uploadsWrapper.web3)
+    ));
+    this.expressApp.use(errorHandling(logger));
```
spherefoundry: To be in line with the ServerWorker implementation, all the express and server construction should be moved to the `work` method, or in this case (we are inheriting PeriodicWorker) into the `beforeWorkLoop` and `afterWorkLoop` methods.

test/workers/hermes_worker.js
```diff
@@ -55,10 +65,11 @@ describe('Hermes Worker', () => {
     mockDataModelEngine.initialiseBundling.resolves(mockResult);
     mockUploadRepository.bundleSizeLimit.resolves(bundleSizeLimit);
     await hermesWorker.beforeWorkLoop();
+    await hermesWorker.work();
```
spherefoundry: You shouldn't call work in tests. It means the worker starts calling the `periodicWork` method rapidly, and we loose the ability to control the execution time of the actual method under test.

danihodovic:
Fixed the review comments, please have another look :)

spherefoundry:
General remark: the health check server for both workers is almost identical. Instead of duplicating it, it could be placed into a separate class and reused in both places (reducing the amount of code in the workers in the process). 

As this is somewhat urgent I would merge this as is, and refactor in a separate PR:D

PR #235, opened 2018-09-14
Implement restrictions on adding/modifying users

vanruch:


src/services/account_access_definitions.js
```diff
@@ -44,6 +44,50 @@ export default class AccountAccessDefinitions {
     return this.ensureHasPermission(address, allPermissions.createEvent);
   }
 
+  async ensureCanRegisterTheAccount(address, newAccount) {
```
Anze1m: Why `ensureCanRegisterTheAccount`? Why not just `ensureCanRegisterAccount `???

src/services/account_access_definitions.js
```diff
@@ -44,6 +44,50 @@ export default class AccountAccessDefinitions {
...
+    const creator = await this.accountRepository.get(address);
+    await this.ensureCanWorkWithAccount(creator, newAccount);
+  }
+
+  async ensureCanModifyTheAccount(address, accountToChange, accountModificationRequest) {
```
Anze1m: Why `ensureCanModifyTheAccount`? Why not just `ensureCanModifyAccount`???

PR #244, opened 2018-10-12
Check is balance is enough in hermes worker

vanruch:


src/workers/hermes_worker.js
```diff
@@ -36,22 +36,26 @@ export default class HermesWorker extends PeriodicWorker {
   }
 
   async periodicWork() {
+    const storagePeriods = this.strategy.storagePeriods();
+    if (!await this.uploadRepository.checkIfEnoughFundsForUpload(storagePeriods)) {
+      this.logger.error('Not sufficient funds.');
```
spherefoundry: `Insufficient funds to perform bundle upload`

test/workers/hermes_worker.js
```diff
@@ -94,6 +102,12 @@ describe('Hermes Worker', () => {
     expect(mockDataModelEngine.uploadNotRegisteredBundles).to.be.calledTwice;
   });
 
+  it('does not bundle if node has no enough funds', async () => {
```
spherefoundry: `does not bundle if insufficient funds`

PR #248, opened 2018-10-19
Check if bundle was uploaded earlier

vanruch:


src/services/data_model_engine.js
```diff
@@ -218,6 +217,10 @@ export default class DataModelEngine {
     await this.entityRepository.storeBundleShelteringExpirationDate(bundleId, expirationDate);
   }
 
+  async updateBundleProofMetadata(bundleId, blockNumber, transactionHash) {
+    return this.entityRepository.storeBundleProofMetadata(bundleId, blockNumber, transactionHash);
+  }
```
Anze1m: I can't see any reason why this call should be delegated to separate method

PR #251, opened 2018-10-22
[PoC] Not persistent challenge storage

vanruch:


src/services/entity_repository.js
```diff
@@ -45,6 +45,9 @@ export default class EntityRepository {
     // for internal use by repository
     await this.db.collection('assets').ensureIndex({'repository.bundleStubId' : 1});
     await this.db.collection('events').ensureIndex({'repository.bundleStubId' : 1});
+
+    // keep failed challenges for 1 day
+    await this.db.collection('rejectedChallenges').ensureIndex({createdAt: 1}, {expireAfterSeconds: 86400});
```
Anze1m: Maybe move `86400` to some descriptive const?

Anze1m: Btw why is this line in `EntityRepository`?

src/services/challenges_repository.js
```diff
@@ -45,4 +46,12 @@ export default class ChallengesRepository {
     }
     return this.challengesWrapper.resolve(challengeId);
   }
+
+  async wasChallengeRejectedRecently(challengeId) {
+    return (await this.db.collection('rejectedChallenges').findOne({challengeId})) !== null;
+  }
+
+  async markChallengeAsRejected(challengeId) {
+    await this.db.collection('rejectedChallenges').insertOne({challengeId, createdAt: Date.now()});
```
Anze1m: `createdAt` sounds like timestamp of challenge creation. `rejectedAt`? `rejectionTimestamp`?

src/services/entity_repository.js
vanruch: Because its an index, I'd put it to config instead of const

src/services/challenges_repository.js
vanruch: It's a mongo document creation time. Seems fine to me

PR #262, opened 2018-11-13
Optional Sentry

conorseabrook:
- Adds Sentry configuration
- Adds Sentry alerts on low and insufficient funds
- Removes old raven client


danihodovic:
LGTM

PR #263, opened 2018-11-14
updated ambrosus-node-contracts to version 0.0.18;

spherefoundry:


src/utils/deployment.js
```diff
@@ -23,7 +23,9 @@ const deployAll = async (web3, privateKey, logger) => {
     },
     validatorSet: {
       owner: deployingAccount.address,
-      initialValidators : [],
+      initialValidators : [
+        '0xc99F38231501Ca5F2e4c31d5b9Df8060B8182Ba5'
+      ],
```
danihodovic: ?

danihodovic: What is src/utils/deployment.js used for?

Anze1m: Dev-deploy script uses it. It can be used while creating local network mock to see node running against real network.

PR #269, opened 2018-11-19
Stop atlas worker when not enough funds to pay for gas

vanruch:
# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective and that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested my changes locally or in the DigitalOcean droplet 
- [x] I have made sure these changes really fix/add new feature as was intended
- [x] I took care of possible corner cases

### DIFF
--- a/src/utils/web3_tools.js
+++ b/src/utils/web3_tools.js
@@ -139,10 +139,11 @@ export async function waitForChainSync(web3, timeoutInSeconds, iterationCallback
 }
 
 export async function getBalance(web3, address) {
-  return new BN(await web3.eth.getBalance(address));
+  return await web3.eth.getBalance(address);
 }
 
-export function maximalGasPrice() {
-  const GAS_LIMIT = '8000000';
-  return new BN(GAS_LIMIT).mul(new Web3().utils.toWei(new BN(config.defaultGasPrice), 'gwei'));
+export async function checkIfEnoughFundsToPayForGas(web3, address) {
+  const maximalGasPrice = new BN(DEFAULT_GAS).mul(web3.utils.toWei(new BN(config.defaultGasPrice), 'gwei'));
+  return new BN(await getBalance(web3, address)).gte(maximalGasPrice);
 }
+
--- a/src/utils/web3_tools.js
+++ b/src/utils/web3_tools.js
@@ -9,6 +9,7 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined 
 
 import Web3 from 'web3';
 import config from '../../config/config';
+import BN from 'bn.js';
 
 export const DEFAULT_GAS = 4700000;
 
@@ -135,4 +136,13 @@ export async function waitForChainSync(web3, timeoutInSeconds, iterationCallback
   while (await isSyncing(web3)) {
     await sleep(timeoutInSeconds);
   }
+}
+
+export async function getBalance(web3, address) {
+  return new BN(await web3.eth.getBalance(address));
+}
+
+export function maximalGasPrice() {
+  const GAS_LIMIT = '8000000';
+  return new BN(GAS_LIMIT).mul(new Web3().utils.toWei(new BN(config.defaultGasPrice), 'gwei'));
 }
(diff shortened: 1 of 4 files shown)


src/utils/web3_tools.js
```diff
@@ -116,3 +117,11 @@ export function link(contract, name, library) {
   contract.bytecode = contract.bytecode.replace(pattern, address);
 }
 
+export async function getBalance(web3, address) {
+  return new BN(await web3.eth.getBalance(address));
```
spherefoundry: Functions should return POJSO as a rule (In this case a string). For context: we already had a lot of trouble when some functions returned a BN and some others have returned a string.

src/utils/web3_tools.js
```diff
@@ -116,3 +117,11 @@ export function link(contract, name, library) {
...
+}
+
+export function maximalGasPrice() {
+  const GAS_LIMIT = '8000000';
+  return new BN(GAS_LIMIT).mul(new Web3().utils.toWei(new BN(config.defaultGasPrice), 'gwei'));
```
spherefoundry: Same here. For calculations it's ok to wrap into BN. But then the result should be converted back into string.

spherefoundry: ⚠️  Magic number warning ⚠️ Why not use the DEFAULT_GAS? All transactions are performed with it by default.

src/workers/atlas_worker.js
```diff
@@ -36,6 +37,10 @@ export default class AtlasWorker extends PeriodicWorker {
   }
 
   async periodicWork() {
+    if (!await this.checkIfEnoughFundsToPayForGas()) {
+      await this.addLog('Not enough funds to pay for gas');
```
spherefoundry: Oh, this is problematic. Basically we will spam MongoDB with this message every second. Totalling at 86400 entries a day until this is resolved. I think the check should be performed on every work loop. But the log should only happen on change enough > not-enough and not-enough > enough.

src/workers/atlas_worker.js
```diff
@@ -57,6 +62,10 @@ export default class AtlasWorker extends PeriodicWorker {
     await this.dataModelEngine.cleanupBundles();
   }
 
+  async checkIfEnoughFundsToPayForGas() {
+    return (await getBalance(this.web3)).gt(maximalGasPrice());
```
spherefoundry: Move to web3 tools?

test/utils/web3_tools.js
```diff
@@ -0,0 +1,65 @@
...
+
+  it('maximalGasPrice returns correct value', () => {
+    expect(maximalGasPrice()).to.equal('23500000000000000');
+  });
+
+  describe('checkIfEnoughFundsToPayForGas', () => {
+    let getBalanceMock;
+
+    before(() => {
+      getBalanceMock = sinon.stub(web3.eth, 'getBalance');
```
vanruch: Will be replaced with mock web3 after merge with #272

src/utils/web3_tools.js
```diff
@@ -136,3 +137,17 @@ export async function waitForChainSync(web3, timeoutInSeconds, iterationCallback
     await sleep(timeoutInSeconds);
   }
 }
+
+export async function getBalance(web3, address) {
+  return await web3.eth.getBalance(address);
+}
+
+export function maximalGasPrice() {
+  return new BN(DEFAULT_GAS).mul(new Web3().utils.toWei(new BN(config.defaultGasPrice), 'gwei'))
```
Anze1m: `new Web3()`? 😮

src/utils/web3_tools.js
```diff
@@ -9,6 +9,7 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
 
 import Web3 from 'web3';
 import config from '../../config/config';
+import BN from 'bn.js';
 
 export const DEFAULT_GAS = 4700000;
```
Anze1m: Why isn't it in config?

vanruch:
@spherefoundry CR please 
image

spherefoundry:
:+1: for the improved tests

PR #271, opened 2018-11-19
Better error messages on http requests

vanruch:


src/utils/https_client.js
```diff
@@ -51,20 +51,21 @@ export default class HttpsClient {
     throw Error(`Invalid protocol ${protocol}`);
   }
 
-  validateIncomingStatusCode(statusCode) {
+  validateIncomingStatusCode(statusCode, body, url) {
+    const errorMsg = `Received code ${statusCode} from ${url} with body: ${JSON.stringify(body, null, 4)}`;
```
spherefoundry: I would skip the body part as we have no way of knowing what will be inside and how much space it will take up.

PR #272, opened 2018-11-19
Wait for parity client to sync

vanruch:


test/utils/web3_tools.js
```diff
@@ -0,0 +1,60 @@
...
+
+  before(async () => {
+    web3 = await createWeb3();
+  });
+
+  describe('isSyncing', () => {
+    let isSyncingStub;
+
+    before(() => {
+      isSyncingStub = sinon.stub(web3.eth, 'isSyncing');
```
spherefoundry: A clean stub (instead of a partial stub) would be nicer.

PR #274, opened 2018-11-22
Atlas worker now picks one challenge at a time for resolution

spherefoundry:
This improves on the way Atlas workers peek challenges to be resolved. 

Instead of querying and checking all the challenges upfront and then just iterating over them (takes 3 + N queries to parity), it now makes the query first (3 queries to parity) and performs the check just before potential resolution attempts. It stops on first successful resolution (due to cooldown being in place, further resolutions doesn't make sense. 

The initial query also takes into account that some challenge candidates may have been already resolved by the current node. Which removes them from the query.

Note: this PR also removes the current cleanup mechanism for bundles as this was implemented in a very inefficient manner.

### DIFF
--- a/src/workers/atlas_worker.js
+++ b/src/workers/atlas_worker.js
@@ -35,26 +35,35 @@ export default class AtlasWorker extends PeriodicWorker {
     return this.dataModelEngine.downloadBundle(bundleId, sheltererId);
   }
 
+  async tryWithChallenge(challenge) {
+    try {
+      if (!await this.strategy.shouldFetchBundle(challenge)) {
+        await this.addLog('Decided not to download bundle', challenge);
+        return false;
+      }
+      const bundle = await this.tryToDownload(challenge);
+      if (!await this.strategy.shouldResolveChallenge(bundle)) {
+        await this.addLog('Challenge resolution cancelled', challenge);
+        return false;
+      }
+      await this.tryToResolve(bundle, challenge);
+      await this.strategy.afterChallengeResolution(bundle);
+      return true;
+    } catch (err) {
+      await this.addLog(`Failed to resolve challenge: ${err.message || err}`, challenge, err.stack);
+      return false;
+    }
+  }
+
   async periodicWork() {
     const challenges = await this.challengesRepository.ongoingChallenges();
+    await this.addLog(`Challenges preselected for resolution: ${challenges.length}`);
     for (const challenge of challenges) {
-      try {
-        if (!await this.strategy.shouldFetchBundle(challenge)) {
-          await this.addLog('Decided not to download bundle', challenge);
-          continue;
-        }
-        const bundle = await this.tryToDownload(challenge);
-        if (!await this.strategy.shouldResolveChallenge(bundle)) {
-          await this.addLog('Challenge resolution cancelled', challenge);
-          continue;
-        }
-        await this.tryToResolve(bundle, challenge);
-        await this.strategy.afterChallengeResolution(bundle);
-      } catch (err) {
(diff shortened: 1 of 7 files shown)


src/workers/atlas_worker.js
```diff
@@ -35,26 +35,35 @@ export default class AtlasWorker extends PeriodicWorker {
     return this.dataModelEngine.downloadBundle(bundleId, sheltererId);
   }
 
+  async tryWithChallenge(challenge) {
```
Anze1m: `proceedWithChallenge`?

test/services/challenges_repository.js
```diff
@@ -21,113 +21,75 @@ describe('Challenges repository', () => {
...
-        {challengeId: '3'}
-      ];
-      const timedOutChallenges = [{challengeId: '3'}];
-
-      it('should remove challenges only if resolution count is not less then creation count', async () => {
-        expect(challengesRepository.filterOutFinished(allChallenges, resolvedChallenges, [])).to.deep.equal([
-          {challengeId: '3', count: 4}
-        ]);
-      });
+    it('should remove challenges that where resolved by the current atlas', async () => {
```
vanruch:
```suggestion
    it('should remove challenges that were resolved by the current atlas', async () => {
```

test/services/challenges_repository.js
```diff
@@ -21,113 +21,75 @@ describe('Challenges repository', () => {
...
+      {challengeId: '7'}
+    ];
 
-      it('should remove single challenges that are resolved or timed out', async () => {
-        expect(challengesRepository.filterOutFinished(allChallenges, resolvedChallenges, timedOutChallenges)).to.deep.equal([
-          {challengeId: '4', count: 1},
-          {challengeId: '5', count: 1}
-        ]);
-      });
+    it('should remove challenges that where fully resolved (resolutions count match start count)', async () => {
```
vanruch:
```suggestion
    it('should remove challenges that were fully resolved (resolutions count match start count)', async () => {
```

test/workers/atlas_worker.js
```diff
@@ -62,56 +62,109 @@ describe('Atlas Worker', () => {
...
+        tryToResolveMock.restore();
+      });
 
-    it('periodic work does not resolve if shouldFetchBundle is false', async () => {
-      shouldResolveChallengeStub.resolves(false);
-      await atlasWorker.periodicWork();
-      expect(dataModelEngineMock.downloadBundle).to.be.called;
-      expect(challengesRepositoryMock.resolveChallenge).to.be.not.called;
+      it('returns false if the strategy disqualifies the challenge', async () => {
+        shouldFetchBundleStub.returns(false);
```
vanruch:
```suggestion
        shouldFetchBundleStub.resolves(false);
```

test/workers/atlas_worker.js
```diff
@@ -62,56 +62,109 @@ describe('Atlas Worker', () => {
...
+        expect(await atlasWorker.tryWithChallenge(challenge1)).to.equal(false);
+        expect(tryToDownloadMock).to.not.have.been.called;
+      });
+
+      it('returns false if download attempt of bundle fails', async () => {
```
vanruch:
```suggestion
      it('returns false if an attempt to download the bundle fails', async () => {
```

test/workers/atlas_worker.js
```diff
@@ -62,56 +62,109 @@ describe('Atlas Worker', () => {
...
+        tryToDownloadMock.throws();
```
vanruch:
```suggestion
        tryToDownloadMock.rejects();
```

PR #278, opened 2018-11-23
Atlas Workers should not retry to solve failed challenge resolution attempts for some time

vanruch:


test/services/failed_challenges_cache.js
```diff
@@ -0,0 +1,68 @@
...
+    failedChallengesCache.rememberFailedChallenge(challengeId1, ttl);
+    clock.tick(3000);
+    failedChallengesCache.rememberFailedChallenge(challengeId2, ttl);
+    expect(failedChallengesCache.failedChallengesEndTime).to.deep.equal({
+      [challengeId1]: now + ttl,
+      [challengeId2]: now + ttl + 3
+    });
+  });
+
+  it('didChallengeFailRecently returns false when challenge is not saved as failed', () => {
```
spherefoundry: You could group the test cases into a `describe` block.

PR #289, opened 2018-11-29
Added worker tasks TTL

vanruch:
# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective and that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested my changes locally or in the DigitalOcean droplet 
- [x] I have made sure these changes really fix/add new feature as was intended
- [x] I took care of possible corner cases


config/config.js
```diff
@@ -46,5 +46,6 @@ module.exports = Object.freeze({
 
   requestSizeLimit: process.env.REQUEST_SIZE_LIMIT || '1mb',
 
-  workerLogsTTLInSeconds: process.env.WORKER_LOGS_TTL || 172800 // 2 days
+  workerLogsTTLInSeconds: process.env.WORKER_LOGS_TTL || 172800, // 2 days
+  workerTasksTTLInSeconds: process.env.WORKER_TAKS_TTL || 172800 // 2 days
```
krzysztof-jelski: spelling: TAKS -> TASK

src/services/worker_task_tracking_repository.js
```diff
@@ -7,25 +7,52 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
+  }
+
+  async isTaskRunning(workType) {
+    return (await this.db.collection('workerTasks').findOne({
+      status: this.statuses.running,
+      workType
+    })) !== null;
+  }
+
+  async tryToBeginWork(workType, taskTimeout) {
```
krzysztof-jelski: Previous implementation had one mongo call, this one has 3. This might be a problem if we expect a lot of worker tasks to be started.

test/services/worker_task_tracking_repository.js
```diff
@@ -13,59 +13,163 @@ import {cleanDatabase, connectToMongo} from '../../src/utils/db_utils';
...
+      expect((await getAllTasks())[0].endTime).to.deep.equal(new Date(now + 2000));
+    });
+
+    it('sets status to SUCCESS when the task has succeeded', async () => {
+      const workId = await storage.tryToBeginWork(exampleTaskType, exampleTaskTimeout);
+      await storage.finishWork(workId, true);
+      expect((await getAllTasks())[0].status).to.equal('SUCCESS');
+    });
+
+    it('sets status to FAILED when the task has succeeded', async () => {
```
krzysztof-jelski: *has failed

src/services/worker_task_tracking_repository.js
```diff
@@ -7,25 +7,52 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
-    } catch (error) {
-      if (error.message.includes(`duplicate key error`)) {
-        throw new Error('Work of this type is currently in progress');
-      }
-      throw new Error(error);
+  statuses = {
+    running: 'RUNNING',
+    timeout: 'TIMEOUT',
+    success: 'SUCCESS',
+    failure: 'FAILURE'
```
krzysztof-jelski: I'm not sure if we can assume the worker to be as kind as to let us know about failure in performing the work. As we're doing this bottom-up (we don't have usages of `finishWork` yet), I'd postpone this until later and now stick to just 3 statuses: running, timeout, finished.

src/services/worker_task_tracking_repository.js
```diff
@@ -7,25 +7,52 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
+      workType,
+      status: this.statuses.running
+    });
+    return insertedId;
   }
 
-  async finishWork(taskId) {
-    await this.db.collection('workerTasks').deleteOne({_id: taskId});
+  async finishWork(taskId, success = true) {
+    const status = success ? this.statuses.success : this.statuses.failure;
```
Anze1m: What can be defined as work failure?

test/services/worker_task_tracking_repository.js
```diff
@@ -13,59 +13,163 @@ import {cleanDatabase, connectToMongo} from '../../src/utils/db_utils';
 import config from '../../config/config';
 
 import WorkerTaskTrackingRepository from '../../src/services/worker_task_tracking_repository';
+import sinon from 'sinon';
 
 const {expect} = chai;
 chai.use(chaiAsPromised);
 
-describe('Worker Log Repository', () => {
+describe('Worker Task Tracking Repository', () => {
```
Anze1m: 👍

vanruch:
After giving it a thought, I guess we don't even need running-success state, as logs give us just enough information about this. However we cannot rely on ttl index, as mongod clears documents once a minute. I'll remove part of functionality

test/services/worker_task_tracking_repository.js
```diff
@@ -13,31 +13,44 @@ import {cleanDatabase, connectToMongo} from '../../src/utils/db_utils';
...
   afterEach(async () => {
     await cleanDatabase(db);
+    clock.restore();
   });
 
   after(async () => {
     client.close();
   });
 
+  it('keeps task start time', async () => {
```
Anze1m:
```suggestion
  it('saves task start time', async () => {
```

test/services/worker_task_tracking_repository.js
```diff
@@ -68,4 +81,11 @@ describe('Worker Log Repository', () => {
     await expect(storage.finishWork(fakeTaskId)).to.be.fulfilled;
     await expect(storage.tryToBeginWork(exampleTaskType)).to.be.rejectedWith('Work of this type is currently in progress');
   });
+
+  it('creates index with correct ttl', async () => {
```
Anze1m: Repository does not create index, migration do. Not sure if this test should be here.

vanruch:
I have some doubts about setting ttl length: we cannot change it after migration has been performed. Maybe we need some way to force migrations?

test/services/worker_task_tracking_repository.js
spherefoundry: ++ @Anze1m. The name of the tests is at best misleading. Also: the only reason this tests pass is because someone run a migration once (cleanDatabase doesn't remove indexes).

So IMO: If the repository assumes that some index is necessary for it to work properly, it should check for its presence.

test/services/worker_task_tracking_repository.js
```diff
@@ -13,31 +13,44 @@ import {cleanDatabase, connectToMongo} from '../../src/utils/db_utils';
...
 
   before(async () => {
     ({db, client} = await connectToMongo(config));
     storage = new WorkerTaskTrackingRepository(db);
+    clock = sinon.useFakeTimers(now);
   });
 
   afterEach(async () => {
     await cleanDatabase(db);
+    clock.restore();
```
spherefoundry: the calls to `useFakeTimers` and `clock.restore` are not symmetric (before vs. afterEach) -> only the first test case will be using the fake timer :(

PR #290, opened 2018-11-30
Move Hermes worker to make use of Worker Task Tracking

vanruch:


src/workers/hermes_worker.js
```diff
@@ -10,22 +10,34 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
+    this.workId = null;
     if (!(this.strategy instanceof HermesUploadStrategy)) {
       throw new Error('A valid strategy must be provided');
     }
   }
 
   async periodicWork() {
-    await this.bundleCandidates();
-    await this.uploadWaitingCandidates();
+    this.workId = await this.workerTaskTrackingRepository.tryToBeginWork(HERMES_BUNDLING_WORK_NAME);
```
Anze1m: Why is this line not inside of any `try` block?

vanruch: It is, only in a parent class. When it fails, there's no point of removing finishWork

src/services/worker_task_tracking_repository.js
```diff
@@ -21,7 +21,7 @@ export default class WorkerTaskTrackingRepository {
       if (error.message.includes(`duplicate key error`)) {
         throw new Error('Work of this type is currently in progress');
       }
-      throw new Error(error);
+      throw error;
```
spherefoundry: 👍

src/workers/hermes_worker.js
```diff
@@ -13,20 +13,21 @@ import healthCheckHandler from '../routes/health_check';
 import PeriodicWorker from './periodic_worker';
 import HermesUploadStrategy from './hermes_strategies/upload_strategy';
 
+const HERMES_BUNDLING_WORK_NAME = 'HermesBundling';
```
spherefoundry: I think the proper term is "work type" isn't it?

src/workers/hermes_worker.js
```diff
@@ -84,7 +90,11 @@ export default class HermesWorker extends PeriodicWorker {
   }
 
   async afterWorkLoop() {
-    await this.server.close();
-    await this.mongoClient.close();
+    try {
+      await this.workerTaskTrackingRepository.finishWork(this.workId);
```
spherefoundry: IMO: this is surplus. I wouldn't include the workId into the state of the worker. Instead I would introduce a local variable `workId` inside periodicWork, and fully handle it there.

src/workers/hermes_worker.js
spherefoundry: It is a bad assumption that this is wrapped somewhere in a try block. All handling of the work should be done inside of this method.

PR #291, opened 2018-11-30
Introducing bundle metadata collection

Anze1m:


src/services/entity_repository.js
```diff
@@ -177,17 +177,18 @@ export default class EntityRepository {
   }
 
   async storeBundle(bundle, storagePeriods) {
-    await this.db.collection('bundles').insertOne({metadata: {storagePeriods}, ...bundle});
+    await this.db.collection('bundles').insertOne({...bundle});
+    await this.db.collection('bundle_metadata').insertOne({bundleId: bundle.bundleId, storagePeriods});
```
vanruch: I'm not sure if this is an issue, but there is a chance that first call is fine and second fails, so we will have bundle without metadata. There are transactions in mongo v4, but I don't think i's worth changing version

vanruch:
Looks fine to me, only missing migrations

src/services/entity_repository.js
conorseabrook: We have deployed mongo v4, the problem is that transactions are only available when using a replica set. A possible solution would be to capture the id of the first insert and delete the record if the second insert fails.

src/migrations/20181203153000_move_out_bundles_metadata.js
```diff
@@ -0,0 +1,38 @@
...
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+// eslint-disable-next-line import/prefer-default-export
+export const up = async (db, config, logger) => {
+  await db.collection('bundles').ensureIndex({bundleId : 1}, {unique: true});
```
vanruch: In earlier migration we currently have `await db.collection('bundles').ensureIndex({bundleId: 1});`. According to docs, ensureIndex will just ignore the unique param. Also it's deprecated. Let's use createIndex from now on

src/migrations/20181203153000_move_out_bundles_metadata.js
```diff
@@ -0,0 +1,38 @@
...
+  await db.collection('bundle_metadata').ensureIndex({bundleId : 1}, {unique: true});
+
+  const extractMetadata = (bundle) => {
+    const {metadata} = bundle;
+    const {bundleId} = bundle;
```
vanruch: Why not merge last 2 lines?

src/migrations/20181203153000_move_out_bundles_metadata.js
```diff
@@ -0,0 +1,38 @@
...
+    const result = {bundleId, ...metadata};
+    return result;
```
vanruch: `return {bundleId, ...metadata};` ?

src/migrations/20181203153000_move_out_bundles_metadata.js
Anze1m: `ensureIndex` is alias for `createIndex`. Switching between these won't change anything.

Anze1m: Maybe we should then delete index from previous migration and then create new, with `unique`?

vanruch: It's not. Create index throws error when same index exists with different parameters. And don't always believe mongo documentation :D

src/migrations/20181203153000_move_out_bundles_metadata.js
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  const performMetadataMigration = async (bundle) => {
+    const metadata = extractMetadata(bundle);
+    try {
+      await db.collection('bundle_metadata').insertOne(metadata);
+    } finally {
+      await db.collection('bundles').updateOne({bundleId: bundle.bundleId}, {$unset: {metadata: 1}});
```
vanruch: There's a small problem with it: if db.collection('bundle_metadata').insertOne(metadata) fails for some reason, we later delete metadata here and it's gone FOREVER

src/services/data_model_engine.js
```diff
@@ -173,12 +173,20 @@ export default class DataModelEngine {
...
+    metadata = pick(metadata, 'bundleId');
+    return {...bundle, metadata};
   }
 
   async getBundleMetadata(bundleId) {
-    const {metadata} = await this.getBundle(bundleId);
-    return {...metadata, bundleId};
+    const metadata = await this.entityRepository.getBundleMetadata(bundleId);
+    if (metadata === null) {
+      throw new NotFoundError(`No bundle metadata for bundleId = ${bundleId} found`);
```
vanruch:
```suggestion
      throw new NotFoundError(`No metadata found for bundleId = ${bundleId}`);
``` may be better (too much bundles for 5 words :))

File: src/migrations/20181203153000_move_out_bundles_metadata.js (hunk 5 above)
Comment (spherefoundry) [reply src/migrations/20181203153000_move_out_bundles_metadata.js]: I agree with @vanruch on this. By doing this update inside of finally, we can potentially loose data if the insert fails.
If the try is here to catch a unique index conflict, we should instead do a `if` with `findOne`. If any of the actions then fails for 'really unexpected reasons' the whole migration will fail, and the operator can debug and fix:D

File: src/migrations/20181203153000_move_out_bundles_metadata.js [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+    }
+  };
+
+  const existingBundlesWithMetadata = await db.collection('bundles').find({metadata: {$exists: true}});
+
+  const migrationsCount = await existingBundlesWithMetadata.count();
+
+  await existingBundlesWithMetadata.forEach(performMetadataMigration);
```
spherefoundry: This is a hidden `await Promise.all(map(.....))` 😛

src/services/data_model_engine.js
```diff
@@ -173,12 +173,20 @@ export default class DataModelEngine {
     if (bundle === null) {
       throw new NotFoundError(`No bundle with id = ${bundleId} found`);
     }
-    return bundle;
+    let metadata = await this.entityRepository.getBundleMetadata(bundleId);
+    if (metadata === null) {
+      return bundle;
+    }
```
spherefoundry: One could argue that `this. getBundleMetadata(bundleId)` could be used here instead.

spherefoundry: Additionally: if there is no metadata for a bundle, should we allow to download it. I would think no.

src/services/entity_repository.js
spherefoundry: I would instead make the method in a way as to handle a second try nicely. That is with the current implementation if the first attempt fail on the second call, then a second attempt will fail on the first because of the unique requirement. So a `findOne and insertOne if not present` pattern would be advised.

PR #295, opened 2018-12-04
Stop on first fail when uploading many bundles.

danihodovic:
Parity RPC servers may be rate limited and sending too HTTP requests may
cause the process to fall under a rate limit. In order to avoid the rate
limits we implement incremental back off which starts at 1s and is
multiplied by 2 for each consecutive HTTP request that is rate limited.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -44,6 +44,7 @@
     "no-useless-constructor": 2,
     "no-dupe-class-members": 2,
     "no-duplicate-imports": 2,
+    "no-constant-condition": "off",
     "import/no-mutable-exports": 2,
     "import/prefer-default-export": 1,
     "import/first": 2,
--- a/src/services/data_model_engine.js
+++ b/src/services/data_model_engine.js
@@ -213,19 +213,17 @@ export default class DataModelEngine {
     await this.entityRepository.discardBundling(bundleStubId);
   }
 
-  async uploadAcceptedBundleCandidates(uploadProgressControls = {success: async () => {}, fail: async () => {}, stopOnFail: false}) {
+  async uploadAcceptedBundleCandidates(uploadProgressCallbacks = {success: async () => {}, fail: async () => {}}) {
     const waitingBundlesMetadata = await this.bundleRepository.findBundlesWaitingForUpload();
     for (const {bundleId, storagePeriods} of waitingBundlesMetadata) {
       try {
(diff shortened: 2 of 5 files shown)


danihodovic:
The scenario where this happens is when there is a bulk of bundles for Hermes to upload that has queued up. Hermes will continue trying to upload bundles, despite 429 rate limits which will make the client fall even deeper into the rate limits. Using the back off we fall under the rate limiting zone and bundle uploads succeed.

test/services/data_model_engine.js
```diff
@@ -1160,6 +1160,36 @@ describe('Data Model Engine', () => {
       expect(result.failed.bundle3.message).to.equal('An error');
       expect(result.failed.bundle3.stack).to.exist;
     });
+
+    it('retries uploads when rate limited', async () => {
```
spherefoundry: This test takes 3 seconds to complete as it actually calls the real bluebird delay function internally.

src/services/data_model_engine.js
```diff
@@ -220,6 +221,24 @@ export default class DataModelEngine {
     return summary;
   }
 
+  async uploadBundle(bundleId, storagePeriods) {
```
spherefoundry: The backoff mechanism should be extracted, generalised, and tested on its own (using sinons fake timers for example). This class should just use this mechanism. The code will be cleaner and easier to test this way.

src/services/data_model_engine.js
```diff
@@ -213,16 +213,19 @@ export default class DataModelEngine {
     await this.entityRepository.discardBundling(bundleStubId);
   }
 
-  async uploadAcceptedBundleCandidates(uploadProgressCallbacks = {success: async () => {}, fail: async () => {}}) {
+  async uploadAcceptedBundleCandidates(uploadProgressControls = {success: async () => {}, fail: async () => {}, stopOnFail: false}) {
```
spherefoundry: IMO: As this method is only used once, and with stopOnFail set to true I would argue we should remove the parameter and assume it's true.

.eslintrc.json
```diff
@@ -44,6 +44,7 @@
     "no-useless-constructor": 2,
     "no-dupe-class-members": 2,
     "no-duplicate-imports": 2,
+    "no-constant-condition": "off",
```
spherefoundry: Is this still required?

PR #299, opened 2018-12-06
constrain event size to 10KB

spherefoundry:


src/utils/validations.js
```diff
@@ -84,6 +84,14 @@ class ValidationAndCasting {
     return this;
   }
 
+  isConstrainedToSize(maxLength, errorMsg = `Object is to large`) {
```
vanruch:
```suggestion
  isConstrainedToSize(maxLength, errorMsg = `Object is too large.`) {
```

test/integration/events/creating.js
```diff
@@ -120,7 +120,7 @@ describe('Events Integrations: Create', () => {
       .and.have.property('status', 403);
   });
 
-  it('returns 403 when trying to create an event with access level hither than own access level', async () => {
```
vanruch: 👍

src/utils/validations.js
Anze1m: We usually do not add dot at the end of an error message.
```suggestion
  isConstrainedToSize(maxLength, errorMsg = `Object is too large`) {
```

PR #306, opened 2018-12-07
refactored procedure of selecting assets and events for bundling;

spherefoundry:


src/services/entity_repository.js
```diff
@@ -115,37 +69,61 @@ export default class EntityRepository {
...
+      if (usedSize + size > bundleSizeInBytesLimit) {
+        return false;
+      }
+      usedSize += size;
+      return true;
+    };
 
-    const discardedEntities = this.discardEntitiesForBundling(orderedEntities, bundleItemsCountLimit, bundleSizeInBytesLimit);
-    await this.unsetEntitiesBundlesStubs(discardedEntities);
+    while (!(nextAsset === null && nextEvent === null) && (selectedAssets.length + selectedEvents.length < bundleItemsCountLimit)) {
```
Anze1m: 🍝😱🍝😱🍝😱

src/services/entity_repository.js
```diff
@@ -46,67 +46,21 @@ export default class EntityRepository {
...
+    }
 
-  async unsetEntitiesBundlesStubs(entities) {
-    const update = {
-      $unset: {
-        'repository.bundleStubId': ''
-      }
-    };
-    await this.updateEntities(entities, update);
+    return asset;
```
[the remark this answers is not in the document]
vanruch: 👍

src/services/entity_repository.js
Anze1m:
```suggestion
    while ((nextAsset !== null || nextEvent !== null) && (selectedAssets.length + selectedEvents.length < bundleItemsCountLimit)) {
```
Imo clearer - as it now explicitly says "Next asset or next event is not null and sum of already selected assets and events is not bigger than `bundleItemsCountLimit `"
Btw it should be `<` or `<=`?

src/services/entity_repository.js
```diff
@@ -115,37 +69,61 @@ export default class EntityRepository {
...
+          break;
+        }
+
+        await this.db.collection('assets').updateOne({assetId: nextAsset.assetId}, updateBundleStubId);
+        selectedAssets.push(nextAsset);
+        nextAsset = await assetsCursor.next();
+      } else {
+        if (!fitsIntoSizeLimit(nextEvent)) {
+          break;
+        }
```
vanruch: This can be extracted out of _if_

PR #310, opened 2018-12-10
Improve event size limit to properly handle non-ascii characters

vanruch:


vanruch:
@Anze1m There're no no-ASCII characters in this case, so there's no point of changing it

Anze1m:
Ok, my bad

PR #311, opened 2018-12-10
asset and event deduplication migration added

Anze1m:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/migrations/20181210103000_deduplicate_assets_and_events.js
+++ b/src/migrations/20181210103000_deduplicate_assets_and_events.js
@@ -17,10 +17,9 @@ const deduplicateEntities = async (db, collectionName, uniqueKey) => {
       }
     },
     {
-      $match:
-        {
-          count: {$gt: 1}
-        }
+      $match: {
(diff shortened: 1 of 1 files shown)


src/migrations/20181210103000_deduplicate_assets_and_events.js
```diff
@@ -0,0 +1,76 @@
...
+    }]
+  );
+  while (await cursor.hasNext()) {
+    const document = await cursor.next();
+    const toRemove = document.dups.slice(1);
+    await db.collection('events').deleteMany({
+      _id: {$in: toRemove}
+    });
+  }
+};
```
vanruch: What's funny: you have duplication of deduplication functions XD. Anyway I think we can live with that in migrations

spherefoundry: IMO: it's a quick fix, so lets fix it:D

PR #312, opened 2018-12-11
Entity bundle split refactoring

spherefoundry:
# Checklist:

- [ ] Documentation updated
- [ ] Automatic tests pass
- [ ] Sanity check passed
- [ ] Self-review passed


dev.env
```diff
@@ -13,3 +13,4 @@ export DEFAULT_STAKE="10000000000000000000000"
 export SENTRY_DSN=""
 export SENTRY_IGNORE_ERRORS_REGEX='^(NotFoundError|ValidationError|JsonValidationError|PermissionError)'
 export LOW_FUNDS_WARNING=1000
+export WORKER_INTERVAL=60
```
Anze1m: Maybe even less? Sth like 30sec?

src/services/bundle_downloader.js
```diff
@@ -7,7 +7,7 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 
-export default class EntityDownloader {
+export default class BundleDownloader {
```
Anze1m: I'd consider fetching bundle metadata from `/bundle/{bundleId}/info` endpoint. It's no longer included in bundle itself, but some data may be useful, like `storagePeriods`

Anze1m: Btw `storagePeriods` is an argument in `storeBundle` method. For now all Atlases have bundle and bundle metadata as follows:
```
{
    bundleId: <bundleId>
    storagePeriods: null
}
```
We should fetch `storagePeriods` or change `storeBundle` method

src/services/bundle_repository.js
```diff
@@ -0,0 +1,59 @@
...
+    if (await this.db.collection('bundle_metadata').findOne({bundleId: bundle.bundleId}) === null) {
+      await this.db.collection('bundle_metadata').insertOne({bundleId: bundle.bundleId, storagePeriods});
+    }
+  }
+
+  async storeBundleProofMetadata(bundleId, proofBlock, timestamp, txHash) {
+    await this.db.collection('bundle_metadata').updateOne({bundleId}, {
+      $set: {
+        bundleTransactionHash: txHash,
+        bundleProofBlock: proofBlock,
```
Anze1m: We should add migration changing all `metadata.proofBlock` to `metadata. bundleProofBlock `

PR #315, opened 2018-12-11
Fixed rare error of dropped collections in-between migrations

vanruch:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/migrations/20181206163000_move_out_bundles_metadata.js
+++ b/src/migrations/20181206163000_move_out_bundles_metadata.js
@@ -7,43 +7,14 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v. 
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 
-const deduplicateBundles = async (db) => {
-  const cursor = await db.collection('bundles').aggregate(
-    [{
-      $group: {
-        _id: '$bundleId',
-        dups: {$addToSet: '$_id'},
-        count: {$sum: 1}
-      }
-    }, {
-      $match: {
-        count: {$gt: 1}
-      }
-    }],
-    {
-      allowDiskUse: true,
-      cursor: {}
-    }
-  );
-  while (await cursor.hasNext()) {
-    const document = await cursor.next();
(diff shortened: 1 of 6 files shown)


Anze1m:
So if in `...bundle...` migration we rely on non-explicit collection existence ensurement, maybe we should do the same in `worker_task_ttl`, where we call `indexExists()`?

vanruch:
`indexExists` throws when called on not existing collection

src/migrations/20181206163000_move_out_bundles_metadata.js
```diff
@@ -41,7 +39,8 @@ export const up = async (db, config, logger) => {
   await deduplicateBundles(db);
   await db.collection('bundle_metadata').createIndex({bundleId : 1}, {unique: true});
 
-  if (await db.collection('bundles').indexExists(['bundleId_1'])) {
+  const collectionExists = (await db.listCollections({name: 'bundles'}).toArray()).length > 0;
+  if (collectionExists && await db.collection('bundles').indexExists(['bundleId_1'])) {
```
spherefoundry: We have a lot of duplication here. I think it's time for some `migration_helpers.js`

src/migrations/20181206163000_move_out_bundles_metadata.js
```diff
@@ -7,44 +7,14 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
...
-  await deduplicateBundles(db);
-  await db.collection('bundle_metadata').createIndex({bundleId : 1}, {unique: true});
+  await deduplicateDocuments(db, 'bundles', 'bundleId');
 
-  if (await db.collection('bundles').indexExists(['bundleId_1'])) {
-    await db.collection('bundles').dropIndex('bundleId_1');
-  }
-  await db.collection('bundles').createIndex({bundleId : 1}, {unique: true});
+  await safeCreateUniqueIndex(db, 'bundle_metadata', 'bundleId');
+  await safeCreateUniqueIndex(db, 'bundles', 'bundleId');
```
spherefoundry: so clean 👍

PR #318, opened 2018-12-12
[WIP] fetching metadata added to bundle downloader

Anze1m:
# Checklist:

- [ ] Documentation updated
- [ ] Automatic tests pass
- [ ] Sanity check passed
- [ ] Self-review passed


src/services/data_model_engine.js
```diff
@@ -226,13 +226,28 @@ export default class DataModelEngine {
...
+      throw new Error('Could not fetch the bundle metadata from the shelterer');
+    }
+
     const bundle = await this.bundleDownloader.downloadBundle(nodeUrl, bundleId);
     if (!bundle) {
       throw new Error('Could not fetch the bundle from the shelterer');
     }
+
     this.bundleBuilder.validateBundle(bundle);
     await this.uploadRepository.verifyBundle(bundle);
```
spherefoundry: IMO: We should also validate the bundle metadata. Could you add a task for this in Asana?

PR #323, opened 2018-12-13
Log on bundle uploads

danihodovic:
# Checklist:

- [x] Documentation updated
- [ ] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

Currently we are in a state of failing bundle uploads without being aware of it.


danihodovic:
Rebase on #322

src/services/data_model_engine.js
```diff
@@ -217,8 +236,18 @@ export default class DataModelEngine {
         await this.entityRepository.storeBundleProofMetadata(waitingBundle.bundleId, blockNumber, timestamp, transactionHash);
         await this.bundleRepository.storeBundleProofMetadata(waitingBundle.bundleId, blockNumber, timestamp, transactionHash);
         summary.ok[waitingBundle.bundleId] = {uploadResult};
+        this.logger.info({
```
spherefoundry: I don't like that the DataModelEngine is now aware of the logger. We managed to avoid this until now, and I would like to keep it this way. As an alternative I would propose the `uploadAcceptedBundleCandidate` method to accept a optional callback that can be used to report progress. It will be also easier to test (which you should also do)

PR #324, opened 2018-12-13
Add process instrumentation

danihodovic:
# Checklist:

- [x] ~Documentation updated~
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

# Changes

I've seen unhandledRejection errors bubble up to the process when
operating Hermes workers in production. This causes a zombie process
which isn't operational, but continues running.

The commit introduces the following changes:

on unhandledRejection or uncaughtException errors that are not handled
and bubble up to the process, do the following:

- exit with a status code of 1 so that the process is restarted by it's
  supervisor in a fresh state (in our case by the Docker daemon)

- log the error

- report the error to Sentry (if Sentry is enabled)

Additionally we enable Sentry for the Hermes and Atlas workers rather
than having it run only on the server.

### DIFF
--- a/src/start_atlas.js
+++ b/src/start_atlas.js
@@ -7,12 +7,12 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v. 
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 
-import {WinstonConsoleLogger} from './utils/loggers';
 import AtlasWorker from './workers/atlas_worker';
 import config from '../config/config';
 import Builder from './builder';
 import {Role} from './services/roles_repository';
 import {waitForChainSync} from './utils/web3_tools';
+import {setup} from './utils/instrument_process';
 
 async function start(logger) {
   const builder = new Builder();
@@ -43,10 +43,4 @@ function loadStrategy(strategyName) {
   return new ChallengeResolutionStrategy();
 }
 
-const logger = new WinstonConsoleLogger();
-
-start(logger)
-  .catch((err) => {
-    logger.error(err);
-    process.exit(1);
-  });
+setup(start);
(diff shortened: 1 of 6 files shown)


danihodovic:
Rebase on #322

src/utils/setup_process.js
```diff
@@ -0,0 +1,40 @@
...
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+import * as Sentry from '@sentry/node';
+import {WinstonConsoleLogger} from './loggers';
+import config from '../../config/config';
+
+const reportAndExit = (logger, message, error) => {
+  Sentry.captureException(error);
```
spherefoundry: I would assume this needs some time to send the error msg to Sentry. Is this synchronous? If not, the request will never go through because of the process.exit(1) just 2 lines below.

conorseabrook: It is not synchronous, but there is a way to flush the buffer. I saw an example in the docs on how to flush the queue and then exit on the completion callback.

danihodovic: @conorseabrook do you have a link?

danihodovic: Found it https://docs.sentry.io/error-reporting/configuration/draining/?platform=javascript thanks ;)

conorseabrook: 👍 I was on my phone or I would have posted.

danihodovic:
I've added tests and added logic for flushing Sentry events.

There is a test missing for `uncaughtExceptions` because Mocha makes this really hard to test. See:
https://stackoverflow.com/questions/9025095/how-can-i-test-uncaught-errors-in-mocha
https://stackoverflow.com/questions/18786325/mocha-unit-test-returns-0-as-exit-code-on-uncaught-exceptions
https://gist.github.com/othiym23/4179636

I've tried `prependOnceListener` and removing listeners for uncaughtException and didn't manage to make it work.

If anyone has any ideas or workarounds please let me know.

PR #325, opened 2018-12-17
Grid fs

spherefoundry:
# Done:
- Storing bundles in GridFS.
- Serving bundles using streaming. 

# Not yet done:

- Streaming bundles directly into GridFS while downloading.

# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed


src/migrations/20181213100000_bundles_to_gridfs.js
```diff
@@ -0,0 +1,59 @@
...
+      if (!inGridFS) {
+        await uploadJSONToGridFSBucket(bundleId, bundle, bundlesBucket);
+      }
+
+      await db.collection('bundles').deleteOne({bundleId});
+      if (!inGridFS) {
+        logger.info(`[${inx}/${totalCount}]: ${bundleId} was uploaded to GridFS and removed from the collection`);
+      } else {
+        logger.info(`[${inx}/${totalCount}]: ${bundleId} was already in GridFS. It was removed from the collection`);
+      }
```
vanruch: If someone has millions of bundles, this will produce shitload of logs. Maybe use some kind of progressbar?

vanruch: https://www.npmjs.com/package/cli-progress

src/migrations/20181213100000_bundles_to_gridfs.js
```diff
@@ -0,0 +1,59 @@
...
+      ++movedCount;
+    }
+    ++inx;
+  }
+
+  logger.info(`Moved ${movedCount} bundles to GridFS. Skipped ${skippedCount} bundles.`);
+
+  await db.collection('bundles').drop();
```
vanruch: What if some bundles were present in 'bundles' collection and not in 'bundles_metadata'?

src/migrations/20181213100000_bundles_to_gridfs.js
spherefoundry: a) Most people will not run the migration directly from the console, but rather inside docker => cli-progress will spam the logs even more
b) Having a list of all migrated bundles in the log that I can check later is IMO a good thing

src/migrations/20181213100000_bundles_to_gridfs.js
spherefoundry: I assumed the bundles_metadata collection is the 'master' collection, and that we don't care about bundles that are not indexed in it. 

What would be a counter-example for this? Do we have a situation right now where we have a bundle, and don't have metadata for it?

src/utils/string_write_stream.js
```diff
@@ -0,0 +1,25 @@
...
+import {Writable} from 'stream';
+
+export default class StringWriteStream extends Writable {
+  constructor(options) {
+    super(options);
+    this.result = '';
+  }
+
+  _write(chunk, encoding, callback) {
+    this.result += chunk;
```
vanruch: According to [docs](https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback_1), chunk is usually a Buffer so concatinating with `+` will not work.
```suggestion
    this.result += chunk.toString();
```

test/utils/db_utils.js
```diff
@@ -57,3 +60,40 @@ describe('mongoObjectSize', () => {
     expect(mongoObjectSize({foo: 1, bar: 3})).to.equal(23);
   });
 });
+
+describe('GridFS helper functions', () => {
+  let db;
+  let client;
+  const exampleJsonObject = {
+    widget: {
```
vanruch:
```suggestion
    fidget: {
```

src/utils/string_write_stream.js
```diff
@@ -0,0 +1,25 @@
...
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+import {Writable} from 'stream';
+
+export default class StringWriteStream extends Writable {
```
vanruch: Miss tests

spherefoundry: ✅

spherefoundry:
@vanruch you can have a second look (please also look at my comments)

test/utils/string_write_stream.js
```diff
@@ -0,0 +1,43 @@
...
+    let callbackStub;
+
+    beforeEach(() => {
+      stream = new StringWriteStream();
+      callbackStub = sinon.stub();
+    });
+
+    afterEach(() => {
+
+    });
```
vanruch: Empty afterEach?

test/utils/string_write_stream.js
```diff
@@ -0,0 +1,43 @@
...
+
+    it('appends the chunk to the internal string', () => {
+      expect(stream.get()).to.equal('');
+      stream._write('abc', null, callbackStub);
+      expect(stream.get()).to.equal('abc');
+      stream._write('12345', null, callbackStub);
```
vanruch: I'd add a test with buffer (just in case)

PR #326, opened 2018-12-18
drop workerTasks without startTime field

Anze1m:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/migrations/20181218161000_remove_workerTasks_without_startTime.js
+++ b/src/migrations/20181218161000_remove_workerTasks_without_startTime.js
@@ -1,0 +1,15 @@
+/*
+Copyright: Ambrosus Technologies GmbH
+Email: tech@ambrosus.com
+
+This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+// eslint-disable-next-line import/prefer-default-export
+export const up = async (db, config, logger) => {
+  await db.collection('workerTasks').deleteMany({startTime: {$exists: false}});
+
+  logger.info(`Removed workerTasks without startTime specified`);
+};


src/migrations/migration_helpers.js
```diff
@@ -45,4 +51,4 @@ const deduplicateDocuments = async (db, collectionName, uniqueKey) => {
   }
 };
 
-export {collectionExists, safeDropIndex, deduplicateDocuments, safeCreateUniqueIndex};
```
krzysztof-jelski: We can stop exporting `safeDropIndex`, as it is used only in migration_helpers now.

src/migrations/20181218115000_worker_task_ttl.js
```diff
@@ -11,6 +11,7 @@ import {safeDropIndex} from './migration_helpers';
 
 // eslint-disable-next-line import/prefer-default-export
 export const up = async (db, {workerTasksTTLInSeconds}, logger) => {
+  await db.collection('workerTasks').remove({startTime: {$exists: false}});
```
spherefoundry:
```suggestion
  await db.collection('workerTasks').deleteMany({startTime: {$exists: false}});
```

src/migrations/20181218154000_worker_index_ttl.js
```diff
@@ -7,11 +7,11 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v.
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 
-import {safeDropIndex} from './migration_helpers';
+import {safeDropCollection} from './migration_helpers';
 
 // eslint-disable-next-line import/prefer-default-export
 export const up = async (db, {workerLogsTTLInSeconds}, logger) => {
-  await safeDropIndex(db, 'workerLogs', 'timestamp_-1');
+  await safeDropCollection(db, 'workerLogs');
```
spherefoundry: IMO: this goes too far.

src/migrations/20181218115000_worker_task_ttl.js
spherefoundry: Also, this line should be a separate migration. Moving a existing migration to the end of the queue, is not a good idea. Other migrations written since, may depend on it being run before them.

PR #327, opened 2018-12-20
Improved the way the upload procedure reports progress.

spherefoundry:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/services/data_model_engine.js
+++ b/src/services/data_model_engine.js
@@ -213,17 +213,16 @@ export default class DataModelEngine {
     await this.entityRepository.discardBundling(bundleStubId);
   }
 
-  async uploadAcceptedBundleCandidates(callbacks = {success: async () => {}, fail: async () => {}}) {
-    const waitingBundles = await this.bundleRepository.findBundlesWaitingForUpload();
-    for (const waitingBundle of waitingBundles) {
-      const {bundleId} = waitingBundle;
+  async uploadAcceptedBundleCandidates(uploadProgressCallbacks = {success: async () => {}, fail: async () => {}}) {
+    const waitingBundlesMetadata = await this.bundleRepository.findBundlesWaitingForUpload();
+    for (const {bundleId, storagePeriods} of waitingBundlesMetadata) {
       try {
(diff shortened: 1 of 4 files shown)


src/services/data_model_engine.js
```diff
@@ -213,23 +213,19 @@ export default class DataModelEngine {
     await this.entityRepository.discardBundling(bundleStubId);
   }
 
-  async uploadAcceptedBundleCandidates() {
+  async uploadAcceptedBundleCandidates(callbacks = {success: async () => {}, fail: async () => {}}) {
```
vanruch: I don't think it's a good idea to create new callbacks as a default parameter. Usually it's handled in a way that if callback is defined => call it

vanruch: Or maybe at least use Function() constructor?

krzysztof-jelski: Suggestion: rename `callbacks` to `uploadProgressCallbacks`

spherefoundry: I like the way it is right now as it shows that the functions can be `async`. Even better would be to use TypeScript 😮

src/services/data_model_engine.js
```diff
@@ -213,23 +213,19 @@ export default class DataModelEngine {
...
     const waitingBundles = await this.bundleRepository.findBundlesWaitingForUpload();
```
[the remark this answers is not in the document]
spherefoundry: We also need storagePeriods :D I have modified the local variables a bit, but reducing the return values is IMO an overkill.

src/services/data_model_engine.js
```diff
@@ -213,23 +213,19 @@ export default class DataModelEngine {
...
-    const summary = {
-      ok: {},
-      failed: {}
-    };
     for (const waitingBundle of waitingBundles) {
```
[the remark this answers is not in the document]
spherefoundry: ✅

src/services/data_model_engine.js
spherefoundry: ✅

PR #328, opened 2018-12-21
added indexes for effective querying of assets and event during bundling

spherefoundry:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed


conorseabrook:
Better to use `createIndex` as `ensureIndex` is now just an alias. ensureIndex was deprecated in 3.0

https://docs.mongodb.com/manual/reference/method/db.collection.ensureIndex

PR #330, opened 2019-01-02
Add Prometheus metrics for Hermes bundling

danihodovic:
Adds metrics for Prometheus bundling for the Hermes worker.

`hermes_bundle_uploads_total` as a counter to track successful uploads.

`hermes_bundle_upload_failures_total` as a counter to track failed
uploads.

### DIFF
--- a/src/workers/hermes_worker.js
+++ b/src/workers/hermes_worker.js
@@ -8,6 +8,8 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined 
 */
 
 import express from 'express';
+import promClient from 'prom-client';
+import prometheusMetricsHandler from '../routes/prometheus_metrics.js';
 import asyncMiddleware from '../middlewares/async_middleware';
 import healthCheckHandler from '../routes/health_check';
 import PeriodicWorker from './periodic_worker';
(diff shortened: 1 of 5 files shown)


src/workers/hermes_worker.js
```diff
@@ -37,6 +39,18 @@ export default class HermesWorker extends PeriodicWorker {
     this.expressApp.get('/health', asyncMiddleware(
       healthCheckHandler(mongoClient, dataModelEngine.identityManager.web3)
     ));
+    const registry = new promClient.Registry();
+    this.expressApp.get('/metrics', prometheusMetricsHandler(registry));
+    this.totalBundlesUploaded = new promClient.Counter({
```
spherefoundry: Again this is something I would love to see inside of a separate `Tracking` service. With a very small interface that hides all the prometheus (and possible other technologies) away behind very simple methods like `markBundleUploadSuccess` and `markBundleUploadFailed`.

test/workers/hermes_worker.js
```diff
@@ -159,8 +160,31 @@ describe('Hermes Worker', () => {
   });
 
   it('health checks', async () => {
-    const {port} = hermesWorker.server.address();
     const {status} = await chai.request(`http://localhost:${port}`).get('/health');
     expect(status).to.eql(200);
   });
+
+  it('records prometheus metrics on successful uploads', async () => {
+    const {text: textBefore} = await chai.request(`http://localhost:${port}`).get('/metrics');
```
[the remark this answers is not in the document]
danihodovic: :+1:

src/workers/hermes_worker.js
danihodovic: What is the benefit of that? We would just be moving complexity to a different file.

PR #331, opened 2019-01-03
Improved performance of finding ongoing challenges;

spherefoundry:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/services/challenges_repository.js
+++ b/src/services/challenges_repository.js
@@ -7,7 +7,6 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v. 
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 
-import {put} from '../utils/dict_utils';
 
 export default class ChallengesRepository {
   constructor(challengesWrapper, configWrapper) {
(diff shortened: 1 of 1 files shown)


src/services/challenges_repository.js
```diff
@@ -50,14 +61,19 @@ export default class ChallengesRepository {
...
+    const filtered = this.filterOutFinishedChallenges(
+      startedChallenges,
+      resolvedChallenges,
+      timedOutChallenges,
+      this.challengesWrapper.defaultAddress
+    );
+    const sorted = this.sortChallenges(
+      filtered
     );
+    return sorted;
```
spherefoundry: This is a little bit artificial (we could just do a straight return), but makes the code more readable, and adding performance measure points easier.

src/services/challenges_repository.js
```diff
@@ -17,23 +16,35 @@ export default class ChallengesRepository {
...
+          acc[challengeId] -= 1;
         }
-        return put(acc, challengeId, acc[challengeId] - 1);
+        return acc;
       },
       startedCount
     );
-    return allChallenges.filter(({challengeId}) => unresolvedCount[challengeId] > 0 && !timedOutSet.has(challengeId));
+    const filtered = allChallenges.filter(({challengeId}) => unresolvedCount[challengeId] > 0 && !timedOutSet.has(challengeId));
+    return filtered;
```
krzysztof-jelski: I would inline `filtered`, but I presume it's just as with `sorted` elsewhere - done to make adding performance measurement easier.

spherefoundry: 👍

PR #337, opened 2019-01-09
log the number of recently failed challenges;

spherefoundry:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/workers/atlas_worker.js
+++ b/src/workers/atlas_worker.js
@@ -97,7 +97,7 @@ export default class AtlasWorker extends PeriodicWorker {
         return;
       }
       const challenges = await this.challengesRepository.ongoingChallenges();
-      const recentlyFailedChallenges = challenges.filter((challenge) => this.failedChallengesCache.failedChallengesEndTime[challenge.challengeId] !== undefined);
+      const recentlyFailedChallenges = challenges.filter(({challengeId}) => this.failedChallengesCache.failedChallengesEndTime[challengeId] !== undefined);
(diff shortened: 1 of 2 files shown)


src/workers/atlas_worker.js
```diff
@@ -97,7 +97,8 @@ export default class AtlasWorker extends PeriodicWorker {
         return;
       }
       const challenges = await this.challengesRepository.ongoingChallenges();
-      await this.addLog(`Challenges preselected for resolution: ${challenges.length}`);
+      const recentlyFailedChallenges = challenges.filter(({challengeId}) => this.failedChallengesCache.failedChallengesEndTime[challengeId] !== undefined);
```
krzysztof-jelski:
```suggestion
      const recentlyFailedChallenges = challenges.filter(({challengeId}) => this.failedChallengesCache.failedChallengesEndTime.hasOwnProperty(challengeId);
```

src/workers/atlas_worker.js
```diff
@@ -97,7 +97,8 @@ export default class AtlasWorker extends PeriodicWorker {
...
+      await this.addLog(`Challenges preselected for resolution: ${challenges.length} (out of ${recentlyFailedChallenges.length} failed recently)`);
```
krzysztof-jelski: If I understand correctly, the message should be different:
```suggestion
      await this.addLog(`Challenges preselected for resolution: ${challenges.length} (out of which ${recentlyFailedChallenges.length} have failed recently)`);
```

PR #338, opened 2019-01-10
downloading bundles directly into GridFS using streams;

spherefoundry:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/services/data_model_engine.js
+++ b/src/services/data_model_engine.js
@@ -254,7 +254,7 @@ export default class DataModelEngine {
       await this.uploadRepository.verifyBundle(bundle);
     } catch (err) {
       await this.bundleRepository.removeBundle(bundleId);
-      throw new Error(err);
+      throw new Error(`Bundle failed to validate: ${err.message || err}`);
     }
 
     await this.bundleRepository.storeBundleProofMetadata(
--- a/src/services/data_model_engine.js
+++ b/src/services/data_model_engine.js
@@ -11,6 +11,7 @@ import {NotFoundError, PermissionError, ValidationError} from '../errors/errors'
 import {getTimestamp} from '../utils/time_utils';
 import {pick, put} from '../utils/dict_utils';
 import allPermissions from '../utils/all_permissions';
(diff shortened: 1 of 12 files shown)


src/services/data_model_engine.js
```diff
@@ -11,6 +11,7 @@ import {NotFoundError, PermissionError, ValidationError} from '../errors/errors'
 import {getTimestamp} from '../utils/time_utils';
 import {pick, put} from '../utils/dict_utils';
 import allPermissions from '../utils/all_permissions';
+const pipeline = require('util').promisify(require('stream').pipeline);
```
vanruch: I don't have anything against it, but if we have imports everywhere, is it fine to use require in that one place?

src/services/data_model_engine.js
```diff
@@ -236,23 +237,29 @@ export default class DataModelEngine {
...
-    if (!bundle) {
+    try {
+      const downloadStream = await this.bundleDownloader.openBundleDownloadStream(nodeUrl, bundleId);
+      const writeStream = await this.bundleRepository.openBundleWriteStream(bundleId, bundleMetadata.storagePeriods);
+      downloadStream.on('error', (err) => {
+        writeStream.abort(err);
+      });
+      await pipeline(downloadStream, writeStream);
+    } catch {
       throw new Error('Could not fetch the bundle from the shelterer');
```
vanruch: This error can be misleading in case exception was thrown anywhere but line 240

src/services/data_model_engine.js
spherefoundry: Sadly this a limitation of import. To use imports and achieve the same thing I would have to split it into three lines:
1) import utli <- would also pollute the context with something I don't need otherwise
2) import the unpromisified version of pipeline <- would also pollute context
3) the the promisification and assign it into a const <- the name of the const can conflict with the import in point 2), so one would have to be reassigned.

IMO: using `require` is just simpler and cleaner in this case

PR #341, opened 2019-01-11
Prometheus metrics for Atlas challenges

danihodovic:
# Checklist:

- [X] Code follows the style guidelines
- ~[X] Documentation updated~
- [X] Automatic tests added/update
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/workers/atlas_worker.js
+++ b/src/workers/atlas_worker.js
@@ -8,6 +8,8 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined 
 */
 
 import express from 'express';
+import promClient from 'prom-client';
+import prometheusMetricsHandler from '../routes/prometheus_metrics.js';
 import asyncMiddleware from '../middlewares/async_middleware';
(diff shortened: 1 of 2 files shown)


src/workers/atlas_worker.js
```diff
@@ -64,23 +82,29 @@ export default class AtlasWorker extends PeriodicWorker {
   async tryWithChallenge(challenge) {
     try {
       if (this.failedChallengesCache.didChallengeFailRecently(challenge.challengeId)) {
+        this.atlasChallengeMetrics.inc({status: atlasChallengeStatus.failedRecently});
```
spherefoundry: I'm not sure if we should track this as this is only a result of the `failed` metric and this will generate a lot of noise.

danihodovic: I'd rather collect as many metrics as possible initially to see what the node is exactly doing.

danihodovic:
Thanks for the review!

PR #342, opened 2019-01-17
Reduce fee queries

krzysztof-jelski:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/builder.js
+++ b/src/builder.js
@@ -93,7 +93,6 @@ class Builder {
       this.uploadActions,
       this.shelteringWrapper,
       this.rolesWrapper,
-      this.feesWrapper,
       this.configWrapper,
       lowFundsWarningAmount,
       Sentry
--- a/src/builder.js
+++ b/src/builder.js
@@ -70,7 +70,7 @@ class Builder {
     this.web3 = web3 || await createWeb3(this.config);
     this.migrator = new Migrator(db, this.config);
     this.identityManager = new IdentityManager(this.web3);
-    const {headContractAddress} = this.config;
+    const {headContractAddress, lowFundsWarningAmount} = this.config;
 
     const defaultAddress = await getDefaultAddress(this.web3);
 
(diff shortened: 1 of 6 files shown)


src/builder.js
```diff
@@ -42,6 +42,7 @@ import ChallengesRepository from './services/challenges_repository';
 import Migrator from './migrations/Migrator';
 import FailedChallengesCache from './services/failed_challenges_cache';
 import WorkerTaskTrackingRepository from './services/worker_task_tracking_repository';
+import * as Sentry from '@sentry/node';
```
vanruch: Why `import * as Sentry` when we always use just `import Sentry from ...`?

src/services/upload_repository.js
```diff
@@ -10,28 +10,26 @@ This Source Code Form is “Incompatible With Secondary Licenses”, as defined
...
   }
 
   async ensureBundleIsUploaded(bundleId, storagePeriods) {
     await this.validateNodeIsOnboardedAsHermes();
-    await this.validateEnoughFundsForUploadFee(storagePeriods);
 
     const previousUploader = await this.shelteringWrapper.getBundleUploader(bundleId);
     if (this.isEmptyAddress(previousUploader)) {
-      return this.receiptWithResult(this.uploadsActions.uploadBundle(bundleId, storagePeriods), 'Bundle has been uploaded');
+      return this.receiptWithResult(this.uploadsActions.uploadBundle(bundleId, storagePeriods, this.lowFundsWarningAmount), 'Bundle has been uploaded');
```
vanruch: `lowFundsWarningAmount` should be passed to the UploadActions' constructor, this shall not work now

src/services/upload_repository.js
```diff
@@ -40,27 +38,35 @@ export default class UploadRepository {
...
+      this.handleLowBalanceWarning(uploadReceipt);
+      return {
+        ...uploadReceipt,
+        uploadResult
+      };
+    } catch (error) {
+      if (error instanceof InsufficientFundsToUploadBundleError) {
+        this.sentry.captureException(error);
+        throw error;
+      }
```
vanruch: I'd move rethrow outside of if, cause if we will catch some other type of error, it will be ignored

test/services/upload_repository.js
```diff
@@ -60,22 +58,34 @@ describe('Upload repository', () => {
...
     });
 
     it('throws if not enough funds', async () => {
-      web3Mock.eth.getBalance.resolves(tooSmallBalance);
+      uploadsActionsMock.uploadBundle = sinon.stub().rejects(new InsufficientFundsToUploadBundleError(fee, tooSmallBalance));
       await expect(uploadRepository.ensureBundleIsUploaded(bundleId, storagePeriods)).to.be.eventually.rejected;
+      expect(sentryMock.captureException).to.be.calledOnce;
+    });
+
+    it('sentry log if low balance', async () => {
```
vanruch:
```suggestion
    it('logs to sentry low balance warning', async () => {
```

PR #344, opened 2019-01-21
Challenge caching

Anze1m:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/services/active_challenges_cache.js
+++ b/src/services/active_challenges_cache.js
@@ -7,7 +7,6 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v. 
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 
-import {pick} from '../utils/dict_utils';
 
 export default class ActiveChallengesCache {
   constructor() {
@@ -15,12 +14,12 @@ export default class ActiveChallengesCache {
   }
 
   get activeChallenges() {
-    return this.sortChronologically(Object.entries(this.activeChallengesDict).map(([key, value]) => ({...value, challengeId: key})));
+    return this.sortChronologically(Object.values(this.activeChallengesDict));
   }
 
   add(challenge) {
     if (!this.has(challenge.challengeId)) {
-      this.activeChallengesDict[challenge.challengeId] = pick(challenge, 'challengeId');
+      this.activeChallengesDict[challenge.challengeId] = challenge;
     }
   }
 
@@ -29,7 +28,7 @@ export default class ActiveChallengesCache {
   }
 
   get(challengeId) {
-    return {...this.activeChallengesDict[challengeId], challengeId};
+    return this.activeChallengesDict[challengeId];
   }
 
   expire(challengeId) {
@@ -46,9 +45,11 @@ export default class ActiveChallengesCache {
   }
 
   applyIncomingChallengeEvents(startedChallenges, resolvedChallenges, timedOutChallenges) {
-    const startedChallengesWithAction = startedChallenges.map((challenge) => this.addAction(challenge, () => this.add(challenge)));
-    const resolvedChallengesWithAction = resolvedChallenges.map((challenge) => this.addAction(challenge, () => this.decreaseActiveCount(challenge.challengeId)));
-    const timedOutChallengesWithAction = timedOutChallenges.map((challenge) => this.addAction(challenge, () => this.expire(challenge.challengeId)));
+    const addAction = (challenge, action) => ({...challenge, action});
+
+    const startedChallengesWithAction = startedChallenges.map((challenge) => addAction(challenge, () => this.add(challenge)));
+    const resolvedChallengesWithAction = resolvedChallenges.map((challenge) => addAction(challenge, () => this.decreaseActiveCount(challenge.challengeId)));
+    const timedOutChallengesWithAction = timedOutChallenges.map((challenge) => addAction(challenge, () => this.expire(challenge.challengeId)));
 
     const challengesWithActionList = this.sortChronologically([...startedChallengesWithAction, ...resolvedChallengesWithAction, ...timedOutChallengesWithAction]);
 
@@ -62,9 +63,5 @@ export default class ActiveChallengesCache {
       }
       return left.logIndex - right.logIndex;
     });
-  }
-
-  addAction(challenge, action) {
-    return {...challenge, action};
   }
 }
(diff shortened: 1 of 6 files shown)


Anze1m:
Instead of caching whole interval at once, cache events one by one. Sort by `block`, sort by `logIndex`, decide what event is that and apply to cached state.

test/services/challenges_repository.js
```diff
@@ -183,38 +140,55 @@ describe('Challenges repository', () => {
...
-        [{challengeId: 'resolved', blockNumber: 3, resolverId: 'someResolverId'}],
-        [{challengeId: 'timedOut', blockNumber: 5}],
-        challengeWrapperMock.defaultAddress
+      expect(challengeWrapperMock.challenges).to.be.calledWith(latestBlock + 1, latestBlock + 3);
+      expect(challengeWrapperMock.resolvedChallenges).to.be.calledWith(latestBlock + 1, latestBlock + 3);
+      expect(challengeWrapperMock.timedOutChallenges).to.be.calledWith(latestBlock + 1, latestBlock + 3);
+      expect(challengesRepository.lastSavedBlock).to.equal(latestBlock + 3);
+    });
+
+    it('adds new challenges to cache, decreases active count on resolved and removes timeouted', async () => {
```
krzysztof-jelski: timeouted -> timed out

test/services/active_challenges_cache.js
```diff
@@ -0,0 +1,213 @@
...
+      {blockNumber: 5, logIndex: 7}
+    ];
+
+    it('in increasing blockNumber and logIndex order', () => {
+      expect(activeChallengesCache.sortChronologically(events)).to.deep.equal([{blockNumber: 5, logIndex: 2}, {blockNumber: 5, logIndex: 7}, {blockNumber: 9, logIndex: 4}]);
+    });
+  });
+
+  describe('Applies multiple changes', () => {
+    it('Applies in correct order', () => {
```
krzysztof-jelski: This test is very hard on the reader. You can't quickly figure out why the expected results are what they are. Consider splitting into a couple of smaller tests.

test/services/challenges_repository.js
```diff
@@ -183,38 +140,55 @@ describe('Challenges repository', () => {
       configWrapperMock = {
         challengeDuration: sinon.stub().resolves(challengeDuration)
       };
-      challengesRepository = new ChallengesRepository(challengeWrapperMock, configWrapperMock);
-      sinon.spy(challengesRepository, 'extractChallengeFromEvent');
-      sinon.stub(challengesRepository, 'filterOutFinishedChallenges').returns(notFinishedEvents);
-      sinon.stub(challengesRepository, 'sortChallenges').returns(sortedEvents);
+      activeChallengesCacheMock = {
```
krzysztof-jelski: I would prefer using the real thing in this test (`new ActiveChallengesCache()`).

src/services/active_challenges_cache.js
```diff
@@ -0,0 +1,92 @@
...
+    if (this.has(challenge.challengeId)) {
+      this.activeChallengesDict[challenge.challengeId].count --;
+      if (this.get(challenge.challengeId).count <= 0) {
+        delete this.activeChallengesDict[challenge.challengeId];
+      }
+    }
+  }
+
+  applyIncomingChallengeEvents(startedChallenges, resolvedChallenges, timedOutChallenges) {
+    const startedChallengesActions = startedChallenges.map((challengeAction) => this.addType(challengeAction, 'issued'));
```
krzysztof-jelski: Store actions instead of strings, it will simplify code.

test/services/active_challenges_cache.js
Anze1m: To be done.

test/services/challenges_repository.js
Anze1m: Why? `ActiveChallengesCache ` is tested separately.

test/services/active_challenges_cache.js
```diff
@@ -0,0 +1,186 @@
...
+
+  beforeEach(() => {
+    activeChallengesCache = new ActiveChallengesCache();
+  });
+
+  function aChallenge() {
+    return exampleChallenge;
+  }
+
+  function aChallengeWith(params) {
```
spherefoundry: you can merge `aChallenge` and `aChallengeWith` methods by using a params parameters with a default of {};

test/services/active_challenges_cache.js
```diff
@@ -0,0 +1,186 @@
...
+  let activeChallengesCache;
+  const exampleChallenge = {
+    challengeId: '0xc0ffee',
+    sheltererId: '0xbeef',
+    bundleId: '0xdeadbeef',
+    count: 12,
+    blockNumber: 1,
+    logIndex: 0
+  };
+  const anotherChallengeId = {challengeId: 'anotherChallengeId'};
```
spherefoundry: The name is somewhat confusing. I would expect a hex string, but it's a dict 🤔

test/services/active_challenges_cache.js
```diff
@@ -0,0 +1,186 @@
...
+    activeChallengesCache.add(aChallengeWith({challengeId: '0xbeef', blockNumber: 3}));
+    activeChallengesCache.add(aChallengeWith({challengeId: '0xdeadface', blockNumber: 2, logIndex: 1}));
+    activeChallengesCache.add(aChallengeWith({challengeId: '0xcafebabe', blockNumber: 2, logIndex: 0}));
+
+    expect(activeChallengesCache.activeChallenges).to.deep.equal([
+      aChallengeWith({challengeId: '0xc0ffee', blockNumber: 1}),
+      aChallengeWith({challengeId: '0xcafebabe', blockNumber: 2, logIndex: 0}),
+      aChallengeWith({challengeId: '0xdeadface', blockNumber: 2, logIndex: 1}),
+      aChallengeWith({challengeId: '0xbeef', blockNumber: 3}),
+      aChallengeWith({challengeId: '0xbaddad', blockNumber: 4})
```
spherefoundry: IMO: there is a lot of repetition here with all the `aChallenge...` calls. Maybe assign them to local variables? This will make the method a little bit longer, but IMO more readable.

src/services/challenges_repository.js
```diff
@@ -46,14 +47,17 @@ export default class ChallengesRepository {
     this.activeChallengesCache.applyIncomingChallengeEvents(startedChallenges, resolvedChallenges, timedOutChallenges);
   }
 
-  async updateBlockInfo() {
+  async getBlockInfo() {
     const challengeDuration = await this.configWrapper.challengeDuration();
```
krzysztof-jelski: We can save one additional call to parity if we did this only in case when there is no lastSavedBlock (rhs of below conditional)

PR #354, opened 2019-02-11
Move bundle version to idData

vanruch:
# Checklist:

- [x] Documentation updated
- [x] Automatic tests pass
- [x] Sanity check passed
- [x] Self-review passed

### DIFF
--- a/src/services/bundle_builder.js
+++ b/src/services/bundle_builder.js
@@ -63,10 +63,12 @@ export default class BundleBuilder {
       this.extractBundleDataNecessaryForValidationFromStream(readStream),
       pipeline(readStream, writeStream)
     ]);
-    if (!this.supportDeprecatedBundleVersions) {
+    if (minimalBundleForLatestVersionValidation.content.idData.version === LATEST_BUNDLE_VERSION) {
       this.validateBundle(minimalBundleForLatestVersionValidation, bundleItemsCountLimit);
-    } else {
+    } else if (this.supportDeprecatedBundleVersions) {
       this.validateBundleWithVersionBefore3(minimalBundleForLatestVersionValidation, bundleItemsCountLimit);
+    } else {
+      throw new ValidationError(`Only supported bundle version is: ${LATEST_BUNDLE_VERSION}`);
     }
   }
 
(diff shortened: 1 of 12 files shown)


src/services/bundle_builder.js
```diff
@@ -51,30 +55,39 @@ export default class BundleBuilder {
...
-  async extractBundleDataNecessaryForValidationFromStream(readableStream, bundleVersion) {
+  async validateStreamedBundle(readStream, writeStream, bundleItemsCountLimit) {
+    readStream.on('error', (err) => {
+      writeStream.abort(err);
+    });
+    const [minimalBundleForLatestVersionValidation] = await Promise.all([
+      this.extractBundleDataNecessaryForValidationFromStream(readStream),
+      pipeline(readStream, writeStream)
+    ]);
+    if (!this.supportDeprecatedBundleVersions) {
```
spherefoundry: This means that even for version 3 bundles, if this config flag is set, then it will always be validated as pre-3. Instead, you should first try to get the version (assume 1 if not pressent) and then only validate against pre-3 if the flag is set.

src/services/bundle_builder.js
```diff
@@ -51,30 +55,41 @@ export default class BundleBuilder {
...
+    const [minimalBundleForLatestVersionValidation] = await Promise.all([
+      this.extractBundleDataNecessaryForValidationFromStream(readStream),
+      pipeline(readStream, writeStream)
+    ]);
+    if (minimalBundleForLatestVersionValidation.content.idData.version === LATEST_BUNDLE_VERSION) {
+      this.validateBundle(minimalBundleForLatestVersionValidation, bundleItemsCountLimit);
+    } else if (this.supportDeprecatedBundleVersions) {
+      this.validateBundleWithVersionBefore3(minimalBundleForLatestVersionValidation, bundleItemsCountLimit);
+    } else {
+      throw new ValidationError(`Only supported bundle version is: ${LATEST_BUNDLE_VERSION}`);
```
spherefoundry:
```suggestion
      throw new ValidationError(`Only bundles with version ${LATEST_BUNDLE_VERSION} are supported`);
```

PR #368, opened 2019-05-16
Change metadata download flow

vanruch:


src/services/upload_repository.js
```diff
@@ -96,14 +96,14 @@ export default class UploadRepository {
     }
   }
 
-  async complementBundleMetadata(bundleMetadata) {
-    const bundleUploadData = await this.uploadsActions.getBundleUploadData(bundleMetadata.bundleId);
+  async composeBundleMetadataFromBlockchain(bundleId) {
```
krzysztof-jelski: Just noted it's a bit strange it's in upload repository. Looks like it's used only on bundle download.

src/services/bundle_repository.js
```diff
@@ -78,14 +80,33 @@ export default class BundleRepository {
     return await downloadJSONFromGridFSBucket(bundleId, this.bundlesBucket);
   }
 
-  async createBundleMetadata(bundleId, storagePeriods, status = BundleStatusStates.unknown) {
+  async createBundleMetadata(bundleId, storagePeriods, status = BundleStatuses.unknown, additionalRepositoryFields = {}) {
```
krzysztof-jelski: Looks like `additionalRepositoryFields` is not used anywhere but in test and can be deleted.

vanruch: Leftovers from adding holdUntil field :)

src/services/bundle_repository.js
```diff
@@ -78,14 +80,33 @@ export default class BundleRepository {
...
     if (await this.db.collection('bundle_metadata').findOne({bundleId}) === null) {
       await this.db.collection('bundle_metadata').insertOne({
-        bundleId, storagePeriods, repository: {status}
+        bundleId, storagePeriods, repository: {...additionalRepositoryFields, status}
       });
     }
   }
 
+  async storeAdditionalMetadata(bundleId, complementaryMetadata) {
+    const requiredFields = [
```
krzysztof-jelski: I imagine those fields could simply come from the initialMetadata. So in the data model engine we could have sth like: `bundle_repository.updateBundleMetadata(bundleId, additionalBundleMetadataFields(initialMetadata, downloadedMetadata))`

src/services/data_model_engine.js
```diff
@@ -254,15 +255,10 @@ export default class DataModelEngine {
...
-      complementedMetadata.bundleId,
-      complementedMetadata.bundleProofBlock,
-      complementedMetadata.bundleUploadTimestamp,
-      complementedMetadata.bundleTransactionHash
-    );
+    await this.bundleRepository.setBundleRepository(bundleId, BundleStatuses.downloaded, {nodeUrl});
+    await this.bundleRepository.storeAdditionalMetadata(bundleId, downloadedMetadata);
 
-    return complementedMetadata;
+    return initialMetadata;
```
krzysztof-jelski: Doesn't make much difference now, but it would be less surprising if we returned the complementedMetadata (meaning initial merged with downloaded).

src/services/upload_repository.js
vanruch: It doesn't communicate with mongo, only takes bundle data from blockchain, so i guess it's ok, may be misleading though

src/services/data_model_engine.js
```diff
@@ -254,15 +255,10 @@ export default class DataModelEngine {
...
     }
 
-    await this.bundleRepository.setBundleRepository(bundleId, BundleStatusStates.downloaded, {nodeUrl});
-    await this.bundleRepository.storeBundleProofMetadata(
-      complementedMetadata.bundleId,
-      complementedMetadata.bundleProofBlock,
-      complementedMetadata.bundleUploadTimestamp,
-      complementedMetadata.bundleTransactionHash
-    );
+    await this.bundleRepository.setBundleRepository(bundleId, BundleStatuses.downloaded, {nodeUrl});
```
Anze1m: A bit confusing -> `bundleRepository.setBundleRepository`

src/services/bundle_repository.js
```diff
@@ -78,14 +80,33 @@ export default class BundleRepository {
...
+      'bundleId',
+      'bundleTransactionHash',
+      'bundleProofBlock',
+      'bundleUploadTimestamp',
+      'storagePeriods',
+      'repository'
+    ];
+    const additionalMetadataFields = pipe(...requiredFields.map((fieldName) =>
```
Anze1m: I'm not sure if this is proper and needed `pipe` usage

src/services/bundle_repository.js
Anze1m: All required? Not too strict? What if only part of them will be downloaded? (I.E. from outdated node)

src/services/bundle_repository.js
```diff
@@ -111,7 +132,7 @@ export default class BundleRepository {
 
   async isBundleSheltered(bundleId) {
     const repository = await this.getBundleRepository(bundleId);
-    return repository ? repository.status === BundleStatusStates.sheltered : false;
+    return repository ? repository.status === BundleStatuses.sheltered : false;
```
Anze1m: Why `false` if there is no `repository` field?

src/services/data_model_engine.js
vanruch: Yeah

src/services/bundle_repository.js
vanruch: It does pick(pick(pick(pick(pick(data))))), exactly wat pipe does

vanruch: If pick was written as fieldname => dict => ... it would look much nicer

src/services/bundle_repository.js
vanruch: Required was a bad name, changed

src/services/bundle_repository.js
vanruch: We don't know if it's sheltered looking only on status, so yes

PR #370, opened 2019-05-17
Add check of recent logs in worker health checker

vanruch:


src/routes/health_check.js
```diff
@@ -23,7 +27,7 @@ const healthCheckHandler = (mongoClient, web3) => async (req, res) => {
     status.web3.connected = false;
   }
 
-  if (!status.mongo.connected || !status.web3.connected) {
+  if (!status.mongo.connected || !status.web3.connected || status.isWorkerActive === false) {
```
Anze1m: Maybe like this?
```suggestion
  if (!status.mongo.connected || !status.web3.connected || !status.isWorkerActive) {
```

src/services/data_model_engine.js
```diff
@@ -281,4 +281,12 @@ export default class DataModelEngine {
...
     return await this.workerLogRepository.getLogs(logsCount);
   }
+
+  async areLogsRecent(maximalLogAgeInSeconds) {
+    const logs = await this.workerLogRepository.getLogs(1);
+    if (logs.length === 0) {
+      return false;
+    }
+    return Date.now() - logs[0].timestamp <= maximalLogAgeInSeconds * 1000;
+  }
```
Anze1m: I'd go for 
```
    if (logs.length > 0) {
       return Date.now() - logs[0].timestamp <= maximalLogAgeInSeconds * 1000;
    }
   return false;
```
Then . works correctly not only for `0`, but also `null` or `undefined` (just in case)

PR #373, opened 2019-05-21
Introduced holdUntil field in bundle metadata repository

vanruch:


src/migrations/20190521104500_add_hold_until_index.js
```diff
@@ -0,0 +1,15 @@
...
+
+This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
+*/
+
+// eslint-disable-next-line import/prefer-default-export
+export const up = async (db, config, logger) => {
+  await db.collection('bundle_metadata').createIndex({'repository.holdUntil': -1});
+
+  logger.info(`Added index to bundle_metadata.repository.holdUntil`);
+};
```
Anze1m: Why wouldn't we check every bundle with blockchain data and assign proper `holdUntil`?
(Also in this scenario we could determine which bundles should be erased)

src/services/data_model_engine.js
```diff
@@ -255,7 +255,7 @@ export default class DataModelEngine {
       throw new Error(`Could not fetch the bundle from the shelterer: ${err.message || err}`);
     }
 
-    await this.bundleRepository.setBundleRepository(bundleId, BundleStatuses.downloaded, {nodeUrl});
+    await this.bundleRepository.setBundleRepository(bundleId, BundleStatuses.downloaded, {nodeUrl, holdUntil: new Date(challengeExpirationTime)});
```
Anze1m: Why do we store `holdUntil` in two places? `metadata` and `repository`.
After successful resolution we update it only in `repository` and leave outdated in `metadata`.

src/services/upload_repository.js
```diff
@@ -81,8 +81,8 @@ export default class UploadRepository {
     return this.shelteringWrapper.isSheltering(bundleId);
   }
 
-  async expirationDate(bundleId) {
-    return this.shelteringWrapper.shelteringExpirationDate(bundleId);
+  async expirationDateInMs(bundleId) {
```
krzysztof-jelski:
```suggestion
  async bundleExpirationDateInMs(bundleId) {
```

src/migrations/20190521104500_add_hold_until_index.js
vanruch: We store holdUntil in metadata already, so it shouldn't be necessary.

src/services/data_model_engine.js
vanruch: :+1: Added migration

src/services/upload_repository.js
vanruch: 👍

PR #382, opened 2019-05-30
Cleanup worker

vanruch:


src/services/data_model_engine.js
```diff
@@ -274,6 +274,14 @@ export default class DataModelEngine {
     await this.bundleRepository.setBundleRepository(bundleId, BundleStatuses.sheltered, {holdUntil: new Date(bundleExpirationDate)});
   }
 
+  /**
+   * @returns {Promise<void>}: removed bundles count
```
krzysztof-jelski: I think it's `Promise<int>` if it returns count.

src/services/bundle_repository.js
```diff
@@ -124,11 +124,32 @@ export default class BundleRepository {
...
+    await this.db.collection('bundle_metadata').updateMany({
+      $or: [
+        {repository: {$exists: false}},
+        {'repository.holdUntil': {$lt: new Date()}}
+      ]
+    }, {$set: {repository: {status: BundleStatuses.cleanup}}});
+  }
+
+  async cleanupBundles() {
+    const cursor = await this.db.collection('bundle_metadata').find({'repository.status': BundleStatuses.cleanup}, {projection: {bundleId: 1, _id: 0}});
```
Anze1m: Maybe also find `expendable` ones? Just in case they weren't properly deleted.

vanruch: Amount of expendables will increase with each iteration. Also expendable status is set after bundle was removed, so realistically I don't see a case when this happens

src/workers/cleanup_worker.js
```diff
@@ -0,0 +1,34 @@
...
+      return;
+    }
+    try {
+      const removedBundlesCount = await this.dataModelEngine.cleanupOutdatedBundles();
+      this.logger.info(`Removed ${removedBundlesCount} outdated bundles`);
+    } finally {
+      await this.workerTaskTrackingRepository.finishWork(workId);
+    }
+  }
+}
```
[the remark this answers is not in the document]
vanruch: Starting up a worker is easy, but including it into atlas worker would increase complexity significantly

src/services/data_model_engine.js
vanruch: 👍

src/services/bundle_repository.js
Anze1m: Makes sense

src/workers/cleanup_worker.js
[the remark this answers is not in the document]
Anze1m: We can somehow, discreetly inject whole cleanup mechanism into atlas/hermes periodic work, while adding another worker potentially makes deployment and maintenance much more difficult. And also will require a lot of changes in ops repo.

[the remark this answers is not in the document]
vanruch: It starts inside start_atlas, so no changes are required

PR #393, opened 2019-07-08
Reordering isTurnToResolve check

AndrewAR2:
Now the check for the possibility of resolving is before downloading the bundle.

# Checklist:

- [+] Code follows the style guidelines
- [+] Documentation updated
- [+] Corner cases evaluated and handled
- [+] Automatic tests added/update
- [ ] Automatic tests pass
- [ ] Sanity check passed
- [+] Self-review passed

### DIFF
--- a/test/workers/atlas_worker.js
+++ b/test/workers/atlas_worker.js
@@ -156,7 +156,6 @@ describe('Atlas Worker', () => {
         shouldFetchBundleStub.resolves(true);
         tryToDownloadMock.resolves(bundleMetadata);
         shouldResolveChallengeStub.returns(true);
-        isTurnToResolveMock.resolves();
         isTurnToResolveMock.returns(true);
         tryToResolveMock.resolves();
       });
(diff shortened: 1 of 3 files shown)


test/workers/atlas_worker.js
```diff
@@ -214,6 +213,12 @@ describe('Atlas Worker', () => {
         expect(await atlasWorker.tryWithChallenge(challenge1)).to.equal(true);
         expect(strategyMock.afterChallengeResolution).to.have.been.calledWith(challenge1);
       });
+
+      it('does not download bundle if not turn to resolve', async () => {
+        isTurnToResolveMock.returns(false);
+        await atlasWorker.tryWithChallenge(challenge1);
+        expect(tryToDownloadMock).to.not.have.been.called;
+      });
```
Anze1m: Maybe also add `expect(tryToResolveMock).to.not.have.been.called;`?

PR #449, opened 2021-11-02
Fix ambrosus node could fail in certain scenarios

vpowler:
* Revert "Fix memory leak (#447)"
* Refactor code to use updated wrappers from the "ambrosus-node-contracts" package
* Update tests

### DIFF
--- a/src/workers/validator_worker.js
+++ b/src/workers/validator_worker.js
@@ -7,7 +7,6 @@ This Source Code Form is subject to the terms of the Mozilla Public License, v. 
 This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.
 */
 import PeriodicWorker from './periodic_worker';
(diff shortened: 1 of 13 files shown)


src/workers/validator_worker.js
```diff
@@ -21,6 +20,7 @@ export default class HermesBundlesValidatorWorker extends PeriodicWorker {
     this.bundleRepository = bundleRepository;
     this.bundleStoreWrapper = bundleStoreWrapper;
     this.shelteringWrapper = shelteringWrapper;
+    this.shelteringContract = null;
```
AndrewAR2, line 23: Is it really necessary?

AndrewAR2 (changes requested):

src/workers/validator_worker.js
vpowler: It was unnecessary.

